PR #60, opened 2019-03-27
Create style guide for documentation

joepetrowski:
Addresses https://github.com/paritytech/substrate-developer-hub/issues/59

The goal of this style guide is to make our documentation more readable by having consistent grammar and usage across our docs.

For example, readers should not see `my_macro` and `my_macro!` depending on who wrote that particular page.

### DIFF
--- a/style-guide.md
+++ b/style-guide.md
@@ -11,7 +11,18 @@ The goal of this guide is to help you achieve these qualities. A consistent styl
 
 When you have a question that is not covered here, refer to the AP style guide. For example, if you want to know if you should spell out "versus" or use "vs.", "vs", or "v.", then search for "AP style versus" and it will [tell you what to do](https://writingexplained.org/ap-style/ap-style-versus).
 
-## Clarity and Consistency
+## Brevity
+
+Before starting to write anything, you should answer the following (example for SRML reference docs):
+
+- Reader: (A runtime developer)
+- Objective: (The reader will understand the functionality, appropriate usage, and warnings associated with this module.)
+- Action: (The reader will implement the appropriate traits and use the functions safely for his or her application.)
+- Impression: (Parity writes clear and useful documentation.)
+
+For every word and sentence that you write, you should ask yourself if it does productive work in achieving the objective, action, and impression on the target reader. If it does not, you need to change it or remove it.
+
+## Clarity
 
 ### Understand before you write
 
@@ -90,13 +101,13 @@ Statements like, "This function is usually used for..." make assumptions about w
 
 ### That vs. Which
 
-"That" and "which" are often used interchangeably in conversational English, but they have different meanings. That is restrictive: you cannot remove it without changing the meaning of the sentence.
+"That" and "which" are often used interchangeably in conversational English, but they have different meanings. "That" is restrictive: you cannot remove it without changing the meaning of the sentence.
 
 > Dispatchable functions that do not have an `origin` parameter in the function signature can only be called by `Root`.
 
 The clause "that do not have an `origin` parameter in the function signature" changes the meaning of the sentence. The sentence would not be true if you removed it.
 
-Which precedes a nonrestrictive clause. Removing it does not change the meaning of the sentence. A nonrestrictive clause is extra information.
+"Which" precedes a nonrestrictive clause. Removing it does not change the meaning of the sentence. A nonrestrictive clause is extra information.
 
 > Dispatchable functions, which are listed in the `Call` enum, always include an `origin` parameter after macro expansion.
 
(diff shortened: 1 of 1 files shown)


nczhu:
This is great, can we also:  

1. Remind people to refer to this substrate style guide: https://wiki.parity.io/Substrate-Style-Guide

2. Gavin/Shawn had a point around being: clear, precise and concise, as reference level documentation handles bns of assets.

3. Minor: can we ask everyone to use relative links, rather than absolute paths, so major restructurings will break less things. e.g.

joepetrowski:
> 1. Remind people to refer to this substrate style guide: https://wiki.parity.io/Substrate-Style-Guide

See line 189, already did

> 2. Gavin/Shawn had a point around being: clear, precise and concise, as reference level documentation handles bns of assets.

I was hoping that the guidance here, especially in the "Clarity" section, would produce this result. But I can add it explicitly.

> 3. Minor: can we ask everyone to use relative links, rather than absolute paths, so major restructurings will break less things. e.g.

Sure

joepetrowski:
I was thinking of adding this to the style guide, although it's not exactly style, but I think helps in achieving concise writing.

Before starting to write anything, you should answer the following (example for SRML reference docs):
**Reader:** (A runtime developer)
**Objective:** (The reader will understand the functionality, appropriate usage, and warnings associated with this module.)
**Action:** (The reader will implement the appropriate traits and use the functions safely for his or her application.)
**Impression:** (Parity writes clear, precise, and concise documentation.)

For every word and sentence that you write, you should ask yourself if it does productive work in achieving the objective, action, and impression on the target reader. If it does not, you need to change it or remove it.

^Let me know if you think that's useful to include or just making it too long

style-guide.md
```diff
@@ -0,0 +1,220 @@
...
+
+Readers can often infer when a "which" clause is restrictive versus unrestrictive. If you want to be precise (you do), always use "that" for a restrictive clause.
+
+### Subject and Pronoun Disagreement
+
+Ensure that pronouns and possessives match the subject they are referring to.
+
+> Incorrect: The process of an account removing their entire holding of an asset.
+
+This is incorrect because "an account" is singular and "their" is plural.
```
shawntabrizi: I thought that this is now acceptable and the politically correct thing to do...

style-guide.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Links
+
+Links to other documents should be relative paths to avoid broken local paths or outdated http links.
+
+> ```
+> [enum](./enum.Call.html)
+> ```
+
+### Markdown Lists
```
shawntabrizi: I am pretty sure this does not matter, but do we want to care about lists using `-` vs `*`

Or putting a space before a list item like:
```
 - there
 - is a
 - space
 - before
 - the `-`
```

If these have no affects on markdown, then I am fine to omit instructions here

style-guide.md
joepetrowski: Sort of. The Washington Post allows singular they, but pretty much every other formal guide rejects it. Without taking sides in the political correctness debate, I think there's usually an elegant sentence reconstruction that avoids the issue entirely.

style-guide.md
joepetrowski: Thought about this, but AFAIK `-` and `*` are always rendered the same way. Can make a call and put it in if people think it's worthwhile.

joepetrowski:
> Looks good to me. A bit long to be honest, but I think all the points are good to call out.
> 
> I wonder if we could also make a checklist template for people to review their own work before asking for others to review.
> 
> BTW, I think we should mention this is specifically for reference level documentation. For other higher level docs, it is actually quite helpful to write it in a more friendly, conversational tone. (at least how I was taught at Microsoft)

Yes, it is long. The idea of a style guide is that you are already a good writer, and you come to the guide when you have a question. Or an editor who is deeply familiar with the guide (right now, that'd be me) enforces the guide before publication.

I think clear, consistent, and concise writing still applies to higher-level docs. It is possible to write in a conversational tone without rambling.

style-guide.md
shawntabrizi: Na, it is just noise. One less thing to complain about in a review

style-guide.md
joepetrowski: Added another way to handle

gavofyork:
Agree. The Economist is a good example of conversational writing but still concise, clear and precise.

shawntabrizi:
👏

nczhu:
+ additional from @bkchr : 

## Examples
We want examples in documentation to be compilable. This is important because: 
- We want to execute our documentation examples as tests with rustdoc
- This keeps examples in sync with code changes
- Readers get more complete examples that are also copy and paste-able. 

Specifically, do not `rust,ignore` nor `nocompile` for code snippets in documentation. 

Here's a [good example](https://crates.parity.io/sr_api_macros/macro.impl_runtime_apis.html) of using examples in documentation.
And here is a [good example](https://hackmd.io/s/Sy6T0usOE) of you should NOT do.  😛

PR #98, opened 2019-06-06
fix grammar

ltfschoen:


shawntabrizi:
Doing this in the source branch

Issue #90, opened 2019-06-04
Create docs for substrate address format


joepetrowski:
Related to deprecation of the wiki. Need up to date docs on address format.
https://github.com/paritytech/wiki/pull/335

Old:

---
title: External Address Format (SS58)
---
SS58 is a simple address format designed for Substrate based chains. There's no problem with using other address formats for a chain, but this serves as a robust default. It is heavily based on Bitcoin's Base-58-check format with a few alterations.	

 The basic idea is a base-58 encoded value which can identify a specific account on the Substrate chain. Different chains have different means of identifying accounts. SS58 is designed to be extensible for this reason.

 ## Basic Format

 The basic format conforms to:	

 ```	
base58encode ( concat ( <address-type>, <address>, <checksum> ) )	
```	

 That is, the concatenated byte series of address type, address and checksum then passed into a base-58 encoder. The `base58encode` function is exactly as [defined](https://en.wikipedia.org/wiki/Base58) in Bitcoin and IPFS, using the same alphabet as both.	

 ## Address Type	

 The `<address-type>` is one or more bytes that describe the precise format of the following bytes.	

 Currently, there exist four valid values:	

 - 00000000b (0) Polkadot Live (SS58 checksum preimage)	
- 00000001b (1) Polkadot Live (AccountId checksum preimage)	
- 00101010b (42) Generic Substrate wildcard (SS58 checksum preimage)	
- 00101011b (43) Generic Substrate wildcard (AccountId checksum preimage)	

 The latter two are "wildcard" addresses that are meant to be equally valid on all Substrate networks that support fixed-length addresses and/or indexed addresses. For production networks, however, a network-specific version may be desirable to help avoid the key-reuse between networks and some of the problems that it can cause. Substrate Node will default to printing keys in address type 42, though alternative Substrate-based node implementations (e.g. Polkadot) may elect to default to some other type.	

 ## Address Formats for Substrate	

 There are 16 different address formats, identified by the length (in bytes) of the total payload (i.e. including the checksum).	

 - 3 bytes: 1 byte account index, 1 byte checksum	
- 4 bytes: 2 byte account index, 1 byte checksum	
- 5 bytes: 2 byte account index, 2 byte checksum	
- 6 bytes: 4 byte account index, 1 byte checksum	
- 7 bytes: 4 byte account index, 2 byte checksum	
- 8 bytes: 4 byte account index, 3 byte checksum	
[...]

PR #108, opened 2019-06-12
SS58 Address

JoshOrndorff:
My attempt at revising the address format docs from the parity wiki. (see issue #90)

Sources:
Ss58Codec implementation https://github.com/paritytech/substrate/blob/cf4a28a7e707190c44e73451e9da38de6bb160ec/core/primitives/src/crypto.rs#L235-L285

ss58hash function https://github.com/paritytech/substrate/blob/cf4a28a7e707190c44e73451e9da38de6bb160ec/core/primitives/src/crypto.rs#L224-L232

PR of most recent overhaul https://github.com/paritytech/substrate/pull/1795


JoshOrndorff:
Thanks for the fixes Joe!
I've fixed the last todo item as well.

I wouldn't mind someone checking this for correctness, but I'm confident enough that I'm willing to merge it if nobody has time to dig in. Might not be too hard to fact check by following the links above.

The merge conflict is a silly mistake I keep making because I forget to build before pushing which means the translation stuff doesn't get updated. I'll be able to easily resolve it.

PR #109, opened 2019-06-13
Improve SEO features

shawntabrizi:
- [x] Fix Open Graph and Twitter Image
- [x] Update title of the homepage
- [x] Only use one H1 tag per page
- [x] Alt description for images


shawntabrizi:
> 👍 One thing I noticed is that the CI already publish the changes to the website, seems other branches will also trigger CI. Should we disable that?

@JoshOrndorff do we have CI bandwidth limits?

JoshOrndorff:
It's the opposite of bandwidth limits. Its that it's actually building (and thus deploying) more than we want. I think I know the solution. On it now.

JoshOrndorff:
The CI issue is fixed in a5f6cfdc585a4fe1b

I confirmed that it works by pushing to `bogus-branch`

Issue #103, opened 2019-06-11
Create Translation Recruitment Page


shawntabrizi:
Docusaurus supports a "Help Translate" button which takes a user to a defined link:

https://docusaurus.io/docs/en/site-config#translationrecruitinglink-string

<img width="566" alt="image" src="https://user-images.githubusercontent.com/1860335/59270692-6e76d800-8c52-11e9-8ead-df2cd29b116a.png">

We should set up instruction page with possibly even a bounty for users who spend enormous efforts to help us keep docs up to date in different languages.

JoshOrndorff:
I couldn't find any more docs than what's linked above

> translationRecruitingLink [string]
>
>URL for the Help Translate tab of language selection when languages besides English are enabled. This can be included you are using translations but does not have to be.

Adding `translationRecruitingLink: 'Help Translate',` to the siteConfig does create the link as shown in the screenshot, but does not point to a reliable URL.

Consider these two examples.
| page I'm on | Link points to |
| --- | --- |
| docs/en/overview/introduction | http://localhost:3000/docs/en/overview/Help%20Translate |
| http://localhost:3000/docs/en/runtime/macros/construct_runtime | http://localhost:3000/docs/en/runtime/macros/Help%20Translate |

shawntabrizi:
You put a fixed link like "https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/contribute.md"

And then they can read instructions on how to help contribute. The link wouldnt be specific per page (but we may be able to hack something together to do that).

Docusaurus links to: https://crowdin.com/project/docusaurus

kaichaosun:
The Crowdin provides a clear workflow for the translation work. An instruction document for translation work could be helpful. Instead of putting the doc link under `Help Translate`, another option is to put the doc under `Docs`, and leave `Help Translate` links to the Crowdin project.

JoshOrndorff:
> You put a fixed link like "https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/contribute.md"

@shawntabrizi You mean that I should/must do that, not that I _did_ do that correct? I definitely know how to create such a page, but I don't know how to get the "Help Translate" link to point to it.

> leave Help Translate links to the Crowdin project.

[...]

PR #113, opened 2019-06-14
Add help translate link and page.

kaichaosun:
Solve https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/103
This help translate page is still in early stage.


kaichaosun:
Hi, @shawntabrizi 
Can you share your opinion about these changes, and maybe your original thoughts around `Create Translation Recruitment Page`?

PR #115, opened 2019-06-14
WIP: Off-chain workers overview

gautamdhameja:
A high-level overview of off-chain workers for introducing the concepts.


gautamdhameja:
@tomusdrw maybe could take a look?

docs/overview/off-chain-workers.md
```diff
@@ -0,0 +1,28 @@
...
+
+To make the off-chain data integration secure and more efficient, Substrate includes off-chain workers. The off-chain worker subsystem allows execution of long-running tasks (web requests, encryption/decryption and signing of data, random number generation, etc.) which would require longer than the block execution time. 
+
+The off-chain workers have their own Wasm execution environment outside of the Substrate runtime. This separation of concerns is to make sure that the block production is not impacted by the long-running tasks. The communication between the off-chain workers and the Substrate runtime is done using `InherentData`. When a task in the off-chain worker completes execution, its result can be submitted back to the runtime using `InherentData`.
```
tomusdrw: > The communication between the off-chain workers and the Substrate runtime is done using `InherentData`. When a task in the off-chain worker completes execution, its result can be submitted back to the runtime using `InherentData`.

However as the off-chain workers are declared in the same code as the runtime, they can easily access on-chain state for their computations. To communicate results back on chain off-chain workers can submit extrinsics (signed and unsigned transactions) that are later included in subsequent blocks.

docs/overview/off-chain-workers.md
```diff
@@ -0,0 +1,28 @@
...
+
+The following diagram depicts how off-chain workers work within the Substrate node:
+
+![Off-chain Workers](/docs/assets/off-chain-workers.png)
+
+Here are some of the advantages of using off-chain workers in comparison to using external oracles:
+
+1. Off-chain workers, being part of the Substrate node, provide more decentralization as compared to a centralized oracle service.
+2. With a sandboxed Wasm execution environment, off-chain workers are fully secure while still providing seamless integration with the runtime using `InherentData`.
+3. No need for maintaining external "glue" services and infrastructure. This reduces maintenance and infrastructure costs for the node operators.
```
tomusdrw: Off-chain workers code is also stored on-chain, which allows off-chain logic updates via standard governance mechanism and seamless upgrades (due to Wasm execution fallback logic)

docs/overview/off-chain-workers.md
```diff
@@ -0,0 +1,28 @@
...
+
+The following APIs are planned to be supported by off-chain workers:
+
+1. Encryption and Decryption
+2. Signing
+3. Local storage - Get and Set
+4. HTTP Requests
```
tomusdrw: + creation of public/private keys (That one might be sort of implicit via encryption/decryiption/signing))
+ source of randomness

JoshOrndorff:
Why was this article removed from devhub? Now that we covered offchain workers in SCL yesterday, I'm getting questions about whether there are other docs available.

gautamdhameja:
@JoshOrndorff What do you mean by "removed from devhub"? 
It seems to be there - https://substrate.dev/docs/en/overview/off-chain-workers

JoshOrndorff:
You're right. Not sure how I got so confused. I'm glad it's there.

PR #119, opened 2019-06-16
Add tutorial/guide for how to add a new module to your runtime

shawntabrizi:
Uses the Contract module as a pretty comprehensive example of all the things you need to do to import a module into your runtime.


docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -0,0 +1,306 @@
...
+
+```rust
+[dependencies.contract]
+default_features = false
+git = 'https://github.com/paritytech/substrate.git'
+package = 'srml-contract'
+rev = '783ca1892892454e05e234cda5f7a2e42a54461e'
+```
+
+You can see that the Contract module has an `std` feature, thus we need to add the `contract/std` feature to the `std` feature of our runtime:
```
JoshOrndorff: > You can see that the Contract module has an `std` feature

You mean here? https://github.com/paritytech/substrate/blob/v1.0/srml/contract/Cargo.toml#L30

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -0,0 +1,306 @@
...
+error: cannot find macro `vec!` in this scope
+   --> /Users/shawntabrizi/.cargo/git/checkouts/substrate-7e08433d4c370a21/783ca18/core/sr-sandbox/src/../without_std.rs:290:24
+    |
+290 |         let mut return_val = vec![0u8; sandbox_primitives::ReturnValue::ENCODED_MAX_SIZE];
+    |                              ^^^
+
+error: aborting due to previous error
+```
+
+If we look closer, it also has another feature: `core`.
```
JoshOrndorff: I see this on master, but not on v1.0 which is what this tutorial is targeting, right?

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -0,0 +1,306 @@
...
+
+// Build native binaries with `std`
+cargo build --release
+```
+
+## Adding the Contract Module
+
+Now that we have successfully imported the Contract module crate, we need to add it to our Runtime.
+
+If you have followed our [other basic tutorials](tutorials/creating-your-first-substrate-chain.md), you would know that we need to implement a `contract::Trait` and also add `Contract: contract::{...}` to our `construct_runtime!` macro.
```
JoshOrndorff:
```diff
- If you have followed our other basic tutorials, you would know that we need ...
+ If you have followed our other basic tutorials, you may remember that we need ...
```

So it doesn't accidentally sound like, "If you had finished your homework, you would already know this"

docs/tutorials/adding-a-module-to-your-runtime.md
JoshOrndorff: v1.0 compiles fine with or without the core feature.

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -0,0 +1,306 @@
...
+To go into a bit more detail here, we see from the documentation that `type Currency` in the Contract module needs to be defined and support the requirements of the trait `Currency`
+
+```rust
+// From the reference documentation, but also can be found in the Contract module code
+type Currency: Currency<Self::AccountId>
+```
+
+Fortunately, the Balances module implements this type, so we can simply reference `Balances` to gain access to it.
+
+Similarly, `type DetermineContractAddress` requires the trait `ContractAddressFor`. The Contract module itself implements a type with this trait in `contract::SimpleAddressDeterminator`, thus can use that implementation to satisfy our `contract::Trait`. At this point, I really recommend you explore the source code of the [Contract module](https://github.com/paritytech/substrate/blob/v1.0/srml/contract/src/lib.rs) if things don't make sense or you want to gain a deeper understanding.
```
JoshOrndorff: thus **we** can use that implementation

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -0,0 +1,306 @@
...
+}
+```
+
+To enable this, we simply need to add `Contract` to the `OnFreeBalanceZero` hook provided by the Balances module:
+
+**`contract-chain/runtime/src/lib.rs`**
+```rust
+impl balances::Trait for Runtime {
+    /// What to do if an account's free balance gets zeroed.
+    type OnFreeBalanceZero = (Contract);
```
JoshOrndorff: Notice this is the capital C Contract that we defined in construct_runtime.

@shawntabrizi Your discretion on this. I tripped on it for a sec, but it was straightforward to fix.
```
help: a type alias with a similar name exists
    |
184 |     type OnFreeBalanceZero = (Contract);
    |     
```

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -0,0 +1,306 @@
...
+```
+
+Note that you can tweak these numbers to your needs, but these are the values set at the genesis configuration of the main Substrate node. You should place that code block into the `testnet_genesis` function, and then in the `GenesisConfig` object, add:
+
+```rust
+GenesisConfig {
+    ...
+    contract: Some(contract_config),
+}
+```
```
JoshOrndorff: Defining `contract_config` variable and then using it is slightly different than how other modules are done in the node template. Learners may not realize that they've done basically the same thing as:

```rust
GenesisConfig {
        contract: Some(ContractConfig {
            transaction_base_fee: 1,
            transaction_byte_fee: 0,
            transfer_fee: 0,
            creation_fee: 0,
            contract_fee: 21,
            call_base_fee: 135,
            create_base_fee: 175,
            gas_price: 1,
            max_depth: 1024,
            block_gas_limit: 10_000_000,
            current_schedule: Default::default(),
        }),
        ...
}
```

I see that this technique is necessary for
```
// IMPORTANT: this should only be enabled on development chains!
contract_config.current_schedule.enable_println = true;
```

I can guess from the name that's something about allowing printing to the console from within contracts? Will I encounter that anywhere else? Maybe worth just mentioning that what we're doing here isn't really that different from how the other modules are doing it?

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -0,0 +1,306 @@
...
+Finally, you are ready to compile your contract capable node with:
+
+```rust
+./scripts/build.sh
+cargo build --release
+```
+
+## Next Steps
+
+We wont actually go into the details of testing the Contract module functionality added to this chain, but if you want to try it out, you can follow our tutorial on [Deploying Your First Smart Contract] (LINK TO BE ADDED).
```
JoshOrndorff: I'm 100% comfortable publishing this before the deploying tutorial is ready. Maybe just say coming soon.

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -0,0 +1,306 @@
...
+
+Make sure to run `./target/release/contract-chain purge-chain --dev` before you start your chain so that the genesis configuration will be initialized for the Contract module.
+
+### Importing and Adding Other SRML Modules
```
JoshOrndorff: How about just "Adding Other SRML Modules". No reason to import without adding.

website/pages/en/tutorials.js
```diff
@@ -137,6 +137,14 @@ class Tutorials extends React.Component {
                 prerequisite: true,
                 href: "https://github.com/substrate-developer-hub/utxo-workshop"
               },
+              {
+                title: "Adding a Module to Your Runtime",
+                text: "This tutorial will teach you how you can add the Contract module or other SRML modules to your Substrate node template.",
```
JoshOrndorff: Pretty verbose description. What do you think about

```diff
- This tutorial will teach you how you can add the Contract module or other SRML modules to your Substrate node template.
+ Add the Contract module or other SRML modules to your Substrate node template.
```

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -0,0 +1,306 @@
+---
+title: "Adding a Module to Your Runtime"
+---
+
+The [Substrate node template](https://github.com/paritytech/substrate/tree/v1.0/node-template) provides you with a minimal working runtime which you can use to quickly get started building your own custom blockchain. However, in the attempts to remain minimal, it does not include most of the modules in the Substrate runtime module library ([SRML](overview/glossary.md#srml-substrate-runtime-module-library)).
+
+This guide will show you how you can add the [Contract module](https://crates.parity.io/srml_contract/index.html) to your runtime in order to allow your blockchain to support Wasm smart contracts. You can follow similar patterns to add additional modules from the SRML to your runtime, however you should note that each module can be a little different in terms of the specific settings needed to import and use it correctly.
+
+## Prerequisites
+
```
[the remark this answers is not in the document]
shawntabrizi: yeah, if you saw my other issue I opened, I want to build a better framework for enabling reusability of setup instructions. Ideally embed one section of a markdown into another.

https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/123

docs/tutorials/adding-a-module-to-your-runtime.md
shawntabrizi: changing to:

```
To enable this, we simply need to add `Contract` type that we defined in the `construct_runtime!` macro to the `OnFreeBalanceZero` hook provided by the Balances module:
```

May be enough to remind the user they need to use the capital Contract

docs/tutorials/adding-a-module-to-your-runtime.md
shawntabrizi: I think we should keep it in since as I mentioned, it is necessary if you actually want to develop contracts on Substrate.

I agree though, a bit of a bummer not to follow the same pattern.

docs/tutorials/adding-a-module-to-your-runtime.md
shawntabrizi: I just put the future link there. Should make it in when we merge the other PR

shawntabrizi:
@JoshOrndorff I updated all the comments except the ContractConfig thing. I agree I wish it could be better, but I think more people will be pretty frustrated trying to try println in the contracts (as our other tutorials teach).

Maybe we should put all the content of that struct explicitly...

PR #122, opened 2019-06-17
Added Substrate Plasma Chain Workshop Tutorial.

SotaWatanabe:
Added a Plasma on Substrate tutorial. It would be great if I could add this content on Substrate Developer Hub page. @JoshOrndorff Here is it :D
FYI: I asked Fredrik to add our Plasma×Substrate tutorial earlier.


shawntabrizi:
@SotaWatanabe not to be a gatekeeper here, but your tutorial does not really teach an end user anything technical about Substrate. This is more of a "product walkthrough".

If you wanted to write a guide on how you wrote the module that might make sense, but in it's current form I don't think this is the right place for it.

That being said, you are featured on our ["users"](https://substrate.dev/en/users.html) page, where we plan to add a GitHub link soon, and you would get to the guide you wrote that way.

SotaWatanabe:
@shawntabrizi That makes sense. I need to know more about the guideline. Thank you for your feedback anyway.

JoshOrndorff:
Since you requested my review, I'll still give it.

I think this is a useful and well-written document. But I agree with @shawntabrizi that substrate developer hub isn't the right home for it.

Issue #112, opened 2019-06-14
Contents in pages need to use <translate> tag


kaichaosun:
Refer to this [doc](https://docusaurus.io/docs/en/next/translation)

PR #126, opened 2019-06-18
Add translate tags for pages

kaichaosun:
Fix https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/112
Changes:
* Delete i18n/en.json, no need to track as the source file
* Use absolute link for recipes
* Add translate tags for pages


shawntabrizi:
@kaichaosun I think there are some translate tags around names like "Substrate" and "StackOverflow", we probably don't need those right?

Also, if you want to not include `i18n/en.json`, you should add it to the .gitignore

kaichaosun:
@shawntabrizi 
* `website/i18n/*` is already ignored, but en.json was added before.
* I start local server, when I try to get the docs of recipes, it redirects to `localhost:3000/recipes`, use full path link can resolve this issue.

PR #132, opened 2019-06-20
Updated frontend developer story

Stefie:
This is just temporary, we'll work on it in more detail once @tbaut is back


Stefie:
notification test @shawntabrizi

PR #134, opened 2019-06-20
allow singular they

joepetrowski:


style-guide.md
```diff
@@ -123,28 +123,15 @@ Readers can often infer when a "which" clause is restrictive versus unrestrictiv
 
 ### Subject and Pronoun Disagreement
 
-Ensure that pronouns and possessives match the subject they are referring to.
+Singular "they" is acceptable as a pronoun. Make sure that there is no possibility of confusion about to what the pronoun refers (subject vs. object).
```
Robbepop:
```suggestion
Singular "they" is accepted as a pronoun. Make sure that there is no possibility of confusion about to what the pronoun refers (subject vs. object).
```

Robbepop: imo sounds better but not too important

Robbepop:
I ll close my PR (https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/128) in favor of yours.

style-guide.md
joepetrowski: Just changed to be active instead of passive, as I should  have done in the first place.

Robbepop: Yep! Sounds really good now.

PR #139, opened 2019-06-22
Schedule deploy at utc midnight.

kaichaosun:
So that approved translations can be deployed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,14 @@ jobs:
 
 workflows:
   version: 2
-  build_and_deploy:
+  commit_deploy:
     jobs:
       - deploy-website:
           filters: *filter-only-source
+  nightly_deploy:
+    triggers:
+      - schedule:
+          cron: "0 0 * * *"
+          filters: *filter-only-source
+    jobs:
+      - deploy-website


.circleci/config.yml
```diff
@@ -35,7 +35,14 @@ jobs:
...
     jobs:
       - deploy-website:
           filters: *filter-only-source
+  nightly_deploy:
+    triggers:
+      - schedule:
+          cron: "0 0 * * *"
+          filters: *filter-only-source
+    jobs:
+      - deploy-website
```
JoshOrndorff: This workflow appears to correctly run at midnight UTC.

.circleci/config.yml
```diff
@@ -35,7 +35,14 @@ jobs:
 
 workflows:
   version: 2
-  build_and_deploy:
+  commit_deploy:
```
JoshOrndorff: I think this workflow will still run on every push. And since there is only one job that is called from both workflows, the scheduled one will be _in addition_ to every push. Please correct me if I'm misunderstanding.

kaichaosun:
@JoshOrndorff 
Yep, you are right. 
The workflow `commit_deploy` will be triggered for every push. 
The workflow `nightly_deploy` will be triggered every midnight which is to keep the translation up to date.
Actually, the two workflows do not affect each other, they simply share the same job which pulls the latest source code, uploads it to Crowdin, downloads translation, then deploy to github pages.
Let me know if it doesn't make sense.

Issue #123, opened 2019-06-17
Consolidate Various Install Instructions


shawntabrizi:
This is a larger vision to try and reduce the footprint of breaking changes to various setup procedures which may discourage a user from using Substrate or other ecosystem projects.

Tangible goals, in order of execution:

- [ ] A single markdown file which describes the instructions to install and use any product
- [ ] A docusaurus plugin to be able to copy either the whole markdown file or specific sub sections into another document, without having to rewrite the content.
- [ ] A docsify plugin to do the same (already works, but we need to add sub-section functionality for optimal use)
- [ ] A rustdocs plugin to do the same (no idea if this is possible)

JoshOrndorff:
Oh nice. I just made a similar comment when reviewing https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/119

I definitely agree with the goal / vision. But I think writing plugins to copy the content to various places is overkill. What if tutorials just start with,

This tutorial assumes you already have a working [substrate developer environment](#) and the link goes to the single markdown document. It could even be a revised version of this page (possibly with details added) https://substrate.dev/docs/en/quickstart/installing-substrate.

I actually think "set up a developer environment" is a better phrasing than "install" since most people don't want to just install and run substrate, they want to hack with it.

shawntabrizi:
I think there is value in not having a user jump to a bunch of different pages to complete a task. Especially I envision that we may only want the user to follow one section of a page for a particular tutorial.

That being said I agree this could be overkill. I probably wont attempt to do this in the near future so maybe we can do your idea as a step one and see how it works.

kaichaosun:
From the experience to use other tools, like [docusaurus](https://docusaurus.io/docs/en), there are rare cases that make the contents duplicated in a lot of places.

A progressive structure for `Docs` will be helpful for users to be more immersed.

Then as Joshy proposed, we can have links in pages for the pre-requirements which depend on another section in another page.

PR #141, opened 2019-06-23
WIP: Tutorial for module in own crate

JoshOrndorff:
Got a draft of this working. Would love some feedback. Then I'll record a video walkthrough on twitch and publish.


shawntabrizi:
@JoshOrndorff do you have this working somewhere? Can you link to the code?

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,133 @@
+---
+title: "Write a Runtime Module in its Own Crate"
+---
+
+In this tutorial you'll write substrate runtime module that lives in its own crate, and include it in a node based on the node-template. If you prefer you can follow along with the [video walkthrough](todo record after initial feedback).
+
+## Setup a Development Environment
+If you haven't already, follow the guide to [install substrate](../getting-started/installing-substrate). You'll need the substrate-up scripts, but not the standard substrate node, so the `--fast` option is what you want here.
```
shawntabrizi:
```suggestion
If you haven't already, follow the guide to [install substrate](getting-started/installing-substrate.md). You'll need the substrate-up scripts, but not the standard substrate node, so the `--fast` option is what you want here.

I think this is the better way to reference other documents. It let's Docusaurus resolve the actual path.
```

shawntabrizi:
Before you continue this, I think really a solution like this should start with some sort of template repository which the user modifies and adds to their runtime, rather than starting from literal scratch.

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,133 @@
...
+
+We do still need the node template, however, to actually run our module.
+
+## Create a Rust Library Crate
+So if our module isn't going to live in the node template, where _is_ it going to live? In a rust library crate. We can create a blank one with
+```bash
+cargo new --lib test_module
+```
+
+If you aren't familiar with library crates or how to use them in other rust projects, it's probably worth reading the [cargo docs](https://doc.rust-lang.org/cargo/guide/creating-a-new-project.html). In the video walkthrough of this tutorial, I briefly describe a simple number doubling program that uses a library crate for the doubling.
```
shawntabrizi: I am not a fan of referencing the video tutorial via the docs. Seems like the docs should live on their own. For example, should the user at this point watch your video or not?

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Make the Crate a Runtime Module
+The cargo command has created the appropriate files and directories for us.
+Let's replace the contents of `lib.rs` in our new crate with the module template from `runtime/template.rs`. (You can also use `substrate-module-new` to create such a file.)
```
shawntabrizi: Wouldn't mention `substrate-module-new` unless you want to use it as the main tool for this step.

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,133 @@
...
+
+Go ahead and compile your new crate with `cargo build`. It won't compile yet, because that template module has some dependencies from substrate itself. The compiler errors give you clues about what you need to include. We can just copy the relevant dependencies from the original node template runtime's `Cargo.toml`. Add these lines to your crate's `Cargo.toml`.
```
shawntabrizi: There is an existing `[dependencies]` tag which you should mention to remove?

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,133 @@
...
+[dependencies.parity-codec]
+default-features = false
+features = ['derive']
+version = '3.5'
+
+[dependencies.support]
+default_features = false
+git = 'https://github.com/paritytech/substrate.git'
+package = 'srml-support'
+rev = 'bb68456966bf9d767651773dbaadd6787bce1884'
```
shawntabrizi: You should mention something about the `rev`, probably change to `<git-commit>` in the code and tell them to update it.

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,133 @@
...
+```toml
+[dependencies.test_module]
+default_features = false
+path = "../../test_module"
+# Optionally specify a version
+# version = "0.1.0"
+```
+
+And remember to update the `std` section/
+```toml
```
shawntabrizi: the toml tag is doing something weird

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,133 @@
...
+
+```toml
+[dependencies.test_module]
+default_features = false
+path = "../../test_module"
+# Optionally specify a version
+# version = "0.1.0"
+```
+
+And remember to update the `std` section/
```
shawntabrizi: This should probably be written a bit more clearly and with explanation.

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,133 @@
...
+}
+```
+
+3. Update `construct_runtime`
+```rust
+TestModule: test_module::{Module, Call, Storage, Event<T>},
+```
+
+## Appreciate your Work
+At this point you have the template module packaged up as it's own crate, and running as part of your `modular-node`. You can now run your chain with `./target/release/modular-chain --dev` and start the [Apps UI](https://polkadot.js.org/apps/#/explorer) to confirm that the module is working as expected. If you aren't familiar with the Apps UI, you will need to navigate to the settings tab, and select 'Local Node'.
```
shawntabrizi: You should probably mention the user needs to run `./scripts/build.sh` and `cargo build --release` and `purge-chain`

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,133 @@
...
+package = 'srml-system'
+rev = 'bb68456966bf9d767651773dbaadd6787bce1884'
+```
+
+At this point we have a runtime module that compiles. It comes with a test, so let's see if that test passes .
+```bash
+cargo test
+```
+
+Turns out we'll need some additional dependencies to make the test pass. You can see this from the error messages, or by looking at the test module itself. Let's add `primitives`, `runtime-primitives`, and `runtime-io` as dev dependencies. The dev dependencies do not need to be listed as part of `std`. TODO is that even proper phrasing?
```
shawntabrizi: Probably should spend a small aside talking about dev dependencies. Why are these ones in dev dependencies and the others are not?

shawntabrizi:
In the end of this, I got a compilation error related to double import of sr-io and stuff

shawntabrizi:
You are missing a step where the user needs to explicitly add:

```
#![cfg_attr(not(feature = "std"), no_std)]
```

docs/tutorials/runtime-module-in-own-crate.md
JoshOrndorff: I've switched it to branch = 'v1.0' like you did in the template and wrote some text about the options.

docs/tutorials/runtime-module-in-own-crate.md
JoshOrndorff: Irrelevant now that I'm using the template

JoshOrndorff:
> You are missing a step where the user needs to explicitly add `#![cfg_attr(not(feature = "std"), no_std)]`

Yes, I was. But I think that's irrelevant now that it's included in your template.

docs/tutorials/runtime-module-in-own-crate.md
JoshOrndorff: This is what I see after `yarn start`. Did you see something different?
image

Maybe it was related to that specific short snippet?

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,167 @@
+---
+title: "Write a Runtime Module in its Own Crate"
+---
+
+In this tutorial you'll write substrate runtime module that lives in its own crate, and include it in a node based on the node-template. If you prefer you can follow along with the [video walkthrough](todo record after initial feedback).
+
+## Setup a Development Environment
+If you haven't already, follow the guide to [install substrate](getting-started/installing-substrate.md). You'll need the substrate-up scripts, but not the standard substrate node, so the `--fast` option is what you want here.
+
+## Create a new Node Template
```
kaichaosun:
```suggestion
## Create your node
```

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,167 @@
...
+We'll begin by creating a brand new node template.
+```bash
+substrate-node-new modular-chain <your name>
+```
+If you've followed other runtime tutorials, you've probably started this way.
+
+In this tutorial, we're not going to write our module directly as part of the node template, but rather as a separate rust crate. This approach provides a few advantages:
```
kaichaosun:
```suggestion
In this tutorial, we're not going to write our module directly as part of the node template, but rather as a separate Rust crate. This approach provides a few advantages:
```

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,167 @@
...
+./target/release/modular-chain purge-chain --dev
+
+# Start the chain
+./target/release/modular-chain --dev
+```
+
+Finally, start the [Apps UI](https://polkadot.js.org/apps/#/explorer) to confirm that the module is working as expected. If you aren't familiar with the Apps UI, you will need to navigate to the settings tab, and select 'Local Node'.
+
+## Customize your Module
+In this tutorial you've learned to create a runtime module as a it's own crate. You're now ready to replace the template module logic with your own logic. If you're not sure what to code, check out one of our runtime module [tutorials](https://substrate.dev/en/tutorials).
```
kaichaosun:
```suggestion
In this tutorial you've learned to create a runtime module as a Rust crate. You're now ready to replace the template module logic with your own logic. If you're not sure what to code, check out one of our runtime modules in [tutorials](https://substrate.dev/en/tutorials).
```

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,167 @@
+---
+title: "Write a Runtime Module in its Own Crate"
```
kaichaosun:
```suggestion
title: "Write a Runtime Module as a Rust crate"
```

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,167 @@
...
+
+In this tutorial, you'll write a Substrate runtime module that lives in its own crate, and include it in a node based on the node-template. If you prefer you can follow along with the [video walkthrough](todo record after initial feedback).
+
+## Setup a Development Environment
+If you haven't already, follow the guide to [install substrate](getting-started/installing-substrate.md). You'll need the substrate-up scripts, but not the standard substrate node, so the `--fast` option is what you want here.
+
+## Create your node
+We'll begin by creating a brand new node template.
+```bash
+substrate-node-new modular-chain <your name>
```
shawntabrizi: If the user follows this, then they need to either replace everything in the node template with `branch = 'v1.0` or they need to replace everything in the module template with `rev = <git-commit>`

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,167 @@
...
+```
+If you've followed other runtime tutorials, you've probably started this way.
+
+In this tutorial, we're not going to write our module directly as part of the node template, but rather as a separate Rust crate. This approach provides a few advantages:
+1. Our module can be easily imported into other nodes in the future.
+2. Our module can be published to places like github or crates.io without publishing our entire node in the same place.
+
+We do still need the node template, however, to actually run our module.
+
+## Clone the Module Template Repository
```
shawntabrizi: I suggest that you be generous with subsections in general. Makes it easier to comb through the docs and link to relevant pieces of information.

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+You should be able to immediately compile this template module with
+```bash
+cargo build --release
+```
+
+But before we jump straight to including our module in a runtime, let's take a look at the `Cargo.toml` file.
+
+After the standard naming and authorship information, we see a few lines about the `std` feature. Because substrate runtime code targets wasm, all the dependencies we use must be able to compile with `no_std`. Our `Cargo.toml` file begins by telling dependencies to only use their `std` feature when this module also uses its `std` feature. You will need to add each dependency you include here.
```
shawntabrizi:
```suggestion
After the standard naming and authorship information, we see a few lines about the `std` feature, which enables use of [the Rust standard libraries](https://doc.rust-lang.org/std/). Because Substrate runtime code targets Wasm, all the dependencies we use must be able to compile with [`no_std`](https://rust-embedded.github.io/book/intro/no-std.html). Our `Cargo.toml` file begins by telling dependencies to only use their `std` feature when this module also uses its `std` feature. You will need to add each dependency you include here.
```

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,167 @@
...
+```toml
+[features]
+default = ['std']
+std = [
+    'parity-codec/std',
+    'support/std',
+    'system/std',
+]
+```
+Our module will depend on low level features such as `system` and `support` from substrate itself. It will also be depended upon by a substrate-based runtime from above. So some care may be needed to ensure version compatibility on both sides. Substrate does not yet have releases published on crates.io, so instead we'll use git-based versioning.
```
shawntabrizi:
```suggestion
Our module will depend on low level Substrate module libraries such as `system` and `support`. It will also be a dependency by the Substrate-based runtime from above. So some care may be needed to ensure version compatibility on both sides. Substrate does not yet have releases published on crates.io, so instead we need to use git-based versioning.
```

shawntabrizi:
Check out the style guide for little nits like capitalization:
https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/blob/source/style-guide.md#when-to-use-preformatted-text

Specifically:

* **S**ubstrate
* **W**asm

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,167 @@
...
+[dependencies.system]
+default_features = false
+git = 'https://github.com/paritytech/substrate.git'
+package = 'srml-system'
+branch = 'v1.0'
+# Could also use a specific revision
+# rev = '<commit hash>'
+```
+
+The final section of the `Cargo.toml` file specifies the dev dependencies, and is not necessary to make the template compile. Remove that final section to confirm this for yourself.
```
shawntabrizi: This section becomes really weird. Rather than have them remove it and get into an error, add it back, etc..., why not simply specify that `dev-dependencies` are those which are simply used by our tests:

https://doc.rust-lang.org/stable/rust-by-example/testing/dev_dependencies.html

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,167 @@
...
+## Add your Runtime Module to your Node
+With our runtime module now compiling and passing tests (well, passing test anyway) we're ready to add it to our node template.
+
+First we need to add our newly-created crate as a dependency in the node runtime's `Cargo.toml`.
+
+```toml
+[dependencies.test_module]
+default_features = false
+path = "../../test_module"
+# Optionally specify a version
```
shawntabrizi: When would they want to do this?

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,167 @@
...
+```toml
+std = [
+  ...
+  'test_module/std',
+]
+```
+
+Next we'll update `runtime/lib.rs` in modular-chain to actually use the new runtime module. The lines that bring in the template module that is coded directly in the node runtime are still there, so we just need to update them in three places.
+
+1. Remove `mod template` as we've already brought that dependency in via `Cargo.toml`
```
shawntabrizi: The user does not need to remove the template module for this to work.

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,167 @@
...
+And run your node with
+```bash
+# Clear the data directory in case you've done this before
+./target/release/modular-chain purge-chain --dev
+
+# Start the chain
+./target/release/modular-chain --dev
+```
+
+Finally, start the [Apps UI](https://polkadot.js.org/apps/#/explorer) to confirm that the module is working as expected. If you aren't familiar with the Apps UI, you will need to navigate to the settings tab, and select 'Local Node'.
```
shawntabrizi: I think the lack of a "better link" in this section means we probably need a short guide on interacting with the Polkadot UI. We should talk to Jaco and team.

shawntabrizi:
My personal feedback is that you have a writing style which feels like a good narration to a video/follow along, but does not work that well (to me) as a standalone text guide.

For example:
* Referencing specific items without explicitly showing them in the text (errors, code, etc...)
* Having the user do "side quests", without really saying it is optional
* Introducing multiple paths/additional parameters without being explicit when they get used

I think our docs should really be much more about stating facts and guiding the user on a specific, fixed path.

But this entire comment is simply my opinion, and should be read as a comment from a reader, not as a specific follow up task or thing I am asking you to change.

docs/tutorials/runtime-module-in-own-crate.md
```diff
@@ -0,0 +1,167 @@
+---
+title: "Write a Runtime Module as a Rust crate"
+---
+
+In this tutorial, you'll write a Substrate runtime module that lives in its own crate, and include it in a node based on the node-template. If you prefer you can follow along with the [video walkthrough](todo record after initial feedback).
+
+## Setup a Development Environment
+If you haven't already, follow the guide to [install substrate](getting-started/installing-substrate.md). You'll need the substrate-up scripts, but not the standard substrate node, so the `--fast` option is what you want here.
```
shawntabrizi: Why not just add the one-liner here?

docs/tutorials/runtime-module-in-own-crate.md
JoshOrndorff: I like learning by doing. I could just explain it instead of show it, but a user in a hurry could also just decide to take my work for it and not reproduce the error. There won't be much left to this beyond
1. substrate-node-new
2. clone template module
3. insert module in runtime
if we omit portions like this.

docs/tutorials/runtime-module-in-own-crate.md
JoshOrndorff: Removed. I guess it only makes sense if we had taught them to publish the module somewhere.

docs/tutorials/runtime-module-in-own-crate.md
JoshOrndorff: I agree that it isn't strictly necessary. But I think it's wise for two reasons.
1. In general it's better to remove excess cruft than leave it around.
2. I'm trying to illustrate how things look _different_ between writing your module as part of your runtime vs writing it as an external crate. Removing that line emphasizes that we do that step differently now. It emphasizes that we do the stuff in cargo.toml _instead_ of `mod template`.

docs/tutorials/runtime-module-in-own-crate.md
JoshOrndorff: Agreed. I was thinking of putting something like that in the "ecosystem tools" section of devhub. I would prefer not to block this PR on that though.

docs/tutorials/runtime-module-in-own-crate.md
JoshOrndorff: For the same reasons we discussed in https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/123 . The install process could (and hopefully will eventually) change. This way we don't have to update them in every single tutorial, but rather in a single central place.

docs/tutorials/runtime-module-in-own-crate.md
JoshOrndorff: Anyway, I've edited to de-emphasize that side exercise.

JoshOrndorff:
I've tried my best to address the remaining concerns. To me it now feels less like "let's explore this and learn together" and more like "follow these steps to achieve a working result".

I recorded this imperfect, but imho still valuable video walkthrough. https://www.twitch.tv/videos/444104764 Thoughts on whether it should go in?

docs/tutorials/runtime-module-in-own-crate.md
shawntabrizi: I still kind of disagree here. You have not for example removed the `template.rs` file from your project, and you say to "update" the definitions rather than simply write new ones.

I think the more common pattern for a user adding/creating a new module is that they would leave everything about their existing node the same, and simply add the new node on top of it.

If you want to make a statement about the `mod` thing you can, but also this is a little bit implied with the use of all the other modules used in the Substrate Node Template (and any minimal familiarity with Rust)

shawntabrizi:
As a follow up, I think it would make sense to add a section at the end where they upload their module to GitHub and then add it to their node as a GitHub dependency rather than simply referring to it by file path. That or they could also go through the simple steps of referencing the `substrate-module-template` repository directly from GitHub.

But this may not be needed... tell me what you feel

JoshOrndorff:
Yes, I also think a section about pushing the module to github and referencing it from there instead of by path is a good idea. Here's a separate issue for it so we can merge this one. https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/151

PR #144, opened 2019-06-24
Restore Gavs writeup.

JoshOrndorff:
But still keep Joshy's specifics of ed25519 and sr25519.

### DIFF
--- a/docs/overview/ss58-address-format.md
+++ b/docs/overview/ss58-address-format.md
@@ -2,11 +2,11 @@
 title: SS58 Address Format
 ---
 
-Most of this content is maintained at [the github wiki](https://github.com/paritytech/substrate/wiki/External-Address-Format-(SS58) and duplicated here for convenience.
-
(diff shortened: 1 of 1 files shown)


docs/overview/ss58-address-format.md
```diff
@@ -1,15 +1,80 @@
...
+
+The `<address-type>` is one or more bytes that describe the precise format of the following bytes.
+
+Currently, there exist four valid values:
+
+- 00000000b (0) Polkadot Live (SS58 checksum preimage)
+- 00000001b (1) Polkadot Live (AccountId checksum preimage)
+- 00101010b (42) Generic Substrate wildcard (SS58 checksum preimage)
+- 00101011b (43) Generic Substrate wildcard (AccountId checksum preimage)
+- 01000000b - 11111111b (64-255) Reserved for future address format extensions.
```
shawntabrizi: I think the goal of the wiki docs here is to allow new projects to claim an index to be used on their chain. So maybe it makes sense to keep the Substrate based indices on this page, and add a link to the wiki with a reference to say something like "all other indices claimed by the community can be found here" or something

PR #147, opened 2019-06-25
Update low-level-data-format.md

kigawas:
Fix typo

### DIFF
--- a/docs/overview/low-level-data-format.md
+++ b/docs/overview/low-level-data-format.md
@@ -26,10 +26,10 @@ A "compact" or general integer encoding is sufficient for encoding large integer
 
 It is encoded with the two least significant bits denoting the mode:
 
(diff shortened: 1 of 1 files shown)


shawntabrizi:
I think it should be `0x` actually right?

`0b` means it should be a binary blob, and thus `0b02` and `0b03` is not valid.

kigawas:
@shawntabrizi 

It's confusing to use __hex__ number like 0x03 (which is actually means 8 bits) to refer two bits 0b11.

kigawas:
Or unify it with 0x00 0x01 0x02 0x03, but still a bit confusing from the context

JoshOrndorff:
@shawntabrizi I agree that `0b02` does not make any sense. But that text does not appear in the PR."

The change was like `0x02` -> `0b10`

shawntabrizi:
Sorry, I somehow read the PR wrong. You all are right

PR #154, opened 2019-06-28
One-pager on storage initialization

JoshOrndorff:
I've sketched out the structure of the document and drafted the first few sections.

Was hoping to make more progress, but I got stuck on a compile issue
```
error[E0063]: missing field `_genesis_phantom_data` in initializer of `genesis_test_runtime::template::GenesisConfig<genesis_test_runtime::Runtime>`
   --> src/chain_spec.rs:100:24
    |
100 |         template: Some(TemplateModuleConfig {
    |                        ^^^^^^^^^^^^^^^^^^^^ missing `_genesis_phantom_data`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0063`.
error: Could not compile `genesis-test`.

To learn more, run the command again with --verbose.
```
Repo to reproduce that error is https://github.com/JoshOrndorff/genesis-test


JoshOrndorff:
I'm pretty happy with this draft, but would really like a critical review before it gets merged.

@thiolliere @nczhu

docs/runtime/initializing-storage.md
```diff
@@ -0,0 +1,140 @@
...
+decl_storage! {
+  trait Store for Module<T: Trait> as TemplateModule {
+    // Points ante'd to play the game
+    pub Ante get(ante) config(): u32 = 5;
+
+    // Minimum raise on each turn
+    pub MinRaise get(min_raise) config(): u32 = 7;
+  }
+}
+```
```
thiolliere: Actually the Hard-Code default value is not working using config. It can be used without config at all (and without adding config to construct_runtime either)

And the behavior is quite surprising.
What does `= 5` means is that the default value of the storage is 5.
Thus the following test is valid:
```diff
diff --git a/srml/example/src/lib.rs b/srml/example/src/lib.rs
index 7e197c13e..1306c43a9 100644
--- a/srml/example/src/lib.rs
+++ b/srml/example/src/lib.rs
@@ -302,6 +302,8 @@ decl_storage! {
 
                // this one uses the default, we'll demonstrate the usage of 'mutate' API.
                Foo get(foo) config(): T::Balance;
+
+               E: Vec<u64> = vec![1, 3];
        }
 }
 
@@ -603,4 +605,15 @@ mod tests {
                        assert_eq!(Example::foo(), 25);
                });
        }
+
+       #[test]
+       fn toto() {
+               with_externalities(&mut new_test_ext(), || {
+                       assert_eq!(E::get(), vec![1, 3]);
+                       E::put(vec![]);
+                       assert_eq!(E::get(), vec![]);
+                       E::kill();
+                       assert_eq!(E::get(), vec![1, 3]);
+               });
+       }
 }
```

thiolliere: Oh but you are also right I actually didn't know: if you use the hard coded default + the config  then it will implement config default this way:
```rust
#[cfg(feature = "std")]
impl <T: Trait> Default for GenesisConfig<T> {
    fn default() -> Self {
        GenesisConfig{dummy: Default::default(),
                      bar: Default::default(),
                      foo: Default::default(),
                      e: <[_]>::into_vec(box [1, 3]),}
    }
}
```

thus writing 
```rust
		t.extend(GenesisConfig::<Test>{
			.. Default::default()
		}.build_storage().unwrap().0);
```
will give the value `[1, 3]` to e
but writing:
```rust
		t.extend(GenesisConfig::<Test>{
			e: Default::default(),
		}.build_storage().unwrap().0);

```
will give the value `Vec<u64>::default` which is `vec![]`

that looks very confusing, also maybe it worth investigate if we want to implement something more straightforward than this. But this is the current behavior.

docs/runtime/initializing-storage.md
```diff
@@ -0,0 +1,140 @@
...
+decl_storage! {
+  trait Store for Module<T: Trait> as TemplateModule {
+    // Points ante'd to play the game
+    pub Ante get(ante) config(): u32;
+
+    // Minimum raise on each turn
+    pub MinRaise get(min_raise) config(): u32;
+  }
+}
+```
```
thiolliere: yes exact, also note that `get(name)` is not mandatory here: if you don't specify a getter then you have to specify the name of the field of the genesis config struct inside config this way: `config(min_raise)`

docs/runtime/initializing-storage.md
```diff
@@ -0,0 +1,140 @@
...
+
+In our example game, the minimum raise must always be twice the ante. But the ante may be configured by the module deployer.
+```rust
+decl_storage! {
+  trait Store for Module<T: Trait> as TemplateModule {
+    // Points ante'd to play the game
+    pub Ante get(ante) config(): u32;
+
+      // Minimum raise on each turn
+      pub MinRaise get(min_raise) build(|config: &GenesisConfig<T>| config.ante * 2): u32;
```
thiolliere: is there an indentation mistake here ?

docs/runtime/initializing-storage.md
```diff
@@ -0,0 +1,140 @@
...
+  }
+}
+```
+> A bug in the v1.0 branch, occasionally causes a compilation error when specifying the genesis configuration this way. Adding an additional field will resolve it.
+>
+> `_genesis_phantom_data: Default::default(),`
```
thiolliere: I'm not sure this comment here makes sense no ?

docs/runtime/initializing-storage.md
```diff
@@ -0,0 +1,140 @@
...
+    // Minimum raise on each turn
+    pub MinRaise get(min_raise) build(|config: &GenesisConfig<T>| config.atom * 3): u32;
+  }
+
+  add_extra_genesis {
+    // The smallest (atomic) number of points worth considering
+    config(atom): u32;
+  }
+}
+```
```
thiolliere: maybe it worth noting also that someone can make use of the field build in add_extra_genesis here no ? see srml/staking for example.

docs/runtime/initializing-storage.md
JoshOrndorff: The context here is
```rust
decl_storage! {
  trait Store for Module<T: Trait> as TemplateModule {
    pub Ante get(ante) config(): u32;
    pub MinRaise get(min_raise) build(|config: &GenesisConfig<T>| config.ante * 2): u32;
  }
}
```
Where `ante` goes in the genesis config, and `min_raise` is calculated with build. My experience testing this was that I needed `chain_spec.rs` to look like
```rust
template: Some(TemplateModuleConfig {
      ante: 5,
      // Woulkdn't compile without this.
      _genesis_phantom_data: Default::default(),
  }),
```

docs/runtime/initializing-storage.md
JoshOrndorff: I've included a link to that part of the staking module and an explanation of what's happening. I think that shows a better use case than doing it in this simple game example which doesn't really need it.

docs/runtime/initializing-storage.md
JoshOrndorff: Oh! I didn't understand how that worked. I've clarified it in the callout a few lines lower.

docs/runtime/initializing-storage.md
JoshOrndorff: Oh interesting! I also didn't realize all of that.

 I'm not sure it's wise to explain so much subtle detail in this article since the target audience is someone building their first few modules. I've added a callout noting that if you kill the storage value, it will return to the default.

docs/runtime/initializing-storage.md
thiolliere: oh ok like because MinRaise is no longer into config it now needs a phantom_data, ok make sense then it worth keeping it

docs/runtime/initializing-storage.md
thiolliere: it has this effect if you kill or if you `take` it.

Issue #118, opened 2019-06-16
Figure out the Versioning Story


shawntabrizi:
We have Substrate v1.0 now, Substrate v2.0 coming soon, and always Substrate latest/master.

This involves some needed magic around reference documentation, search indexing, clarity to the user, etc...

While it is probably not super important at the time of writing this issue, it is something that we should think about in the long term.

**Task Updated:**
- [x] Enable versioning for source and translated doc.
- [ ] Fix search between different version.
- [ ] Correct links to reference doc in different versions.

kaichaosun:
The translations may also need version management in the future, I haven't seen any use cases for now, let's keep this thread to discuss and track it.

JoshOrndorff:
For reference docusaurus docs on versioning https://docusaurus.io/docs/en/versioning

@kaichaosun There is a [short paragraph](https://docusaurus.io/docs/en/versioning#versioning-and-translations) about versioning with translation at the end.

We can restrict search results to current version [like we did with current language](https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/commit/f07250879d6917887df24dd5aeda6ea6f0a9e030).
But I will say, that restriction seems to be interfering with our search results for non-docusaurus portions of the site.

JoshOrndorff:
Further evidence that we're ready to start versioning the docs. https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/156

JoshOrndorff:
Docusaurus has a nice [versioning fallback feature](https://docusaurus.io/docs/en/versioning#fallback-functionality) such that if you request version 3.0 of a document, but that doc hasn't changed since 1.0, it will automatically serve the 1.0 version of the doc. The only problem is that we're in a situation where we don't yet have full documentation coverage.

**Proposal**
We cut version 1.0.0 soonish (like maybe end of July 2019). This will essentially copy the current docs into a folder called `docs/1.0.0`. From then until we decide otherwise, we write docs in a v1-first fashion:

* Changes specific to v1.0 go in the `docs/1.0.0` folder
* Changes specific to master go directly in the `docs` directory
* Additions go in _both_ directories
* Additions that _do not_ apply to both branches _still_ go in both directories. In this case you have two sub-options.
  * Write two versions (one for master and one for v1)
  * Write against v1, and open an issue to update the v1 version to match master.

At some point in the future we decide to switch to a master-first style.

[...]

PR #157, opened 2019-07-01
Add versioning to devhub

JoshOrndorff:
This is a first draft attempt at adding versioning to devhub following the [docusaurus instructions](https://docusaurus.io/docs/en/versioning).

Remaining tasks / discussions to have before merging
* [x] How does [versioning interact with translations](https://docusaurus.io/docs/en/versioning#versioning-and-translations)
* [ ] How to version content outside of docusaurus (like recipes, kitties, etc)
* [x] What is the process for adding missing docs. Here is [a proposal](https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/118#issuecomment-507384480)


kaichaosun:
Support version in translations is straightforward, the doc details [here](https://docusaurus.io/docs/en/translation#versioned-translations).

website/versioned_docs/version-1.0.0/contracts/cargo-contract.md
```diff
@@ -0,0 +1,143 @@
+---
+title: Cargo Contract: The ink! CLI
```
kaichaosun: The versioned doc has removed the quotes, it caused the related translation doc to be:
```
---
title:
  Cargo Contract: The ink! CLI
id: version-1.0.0-cargo-contract
original_id: cargo-contract
---
```

kaichaosun: @JoshOrndorff and I have a discussion, and we decided to add the quotes to solve it.

JoshOrndorff: I now realize that cutting a version is dropping the quotes from titles in _all_ cases. We only notice it in the ones that have colons (`:`). Maybe a better solution is to not put colons in the title?

kaichaosun:
There is one issue when I'm testing:
* The translation for doc title is not working in the versioned (v1.0.0) doc.
* With `yarn start`, If the title in the translated doc is not the same with the title in source doc, 
Chrome can't access the translation doc,
Firefox fallback to use the title in latest doc, the doc is accessible. 

This is not a blocker for us to process the work of version enable.
* I will test the GitHub page when it's merged, to see if there is the same issue.
* I will do a walk around fix to make sure the title in translated docs is same with the source doc.

I suggest that:
* Agree and fix the colons `:` in the title of source docs.
* Refresh this PR to cut a new v1.0.0 since there are other changes happened in recent a few days.
* Once add or modify docs, we need to make sure both latest and v1.0.0 are treated thoroughly once this PR gets merged.

Are there any other concerns from @JoshOrndorff @shawntabrizi ?

kaichaosun:
I updated the source doc's title in this [commit](https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/157/commits/6288af7033ef2f8732e4dfaf4bb4d2d14cd48f39).
@JoshOrndorff Are you able to cut the new v1.0.0 version with master changes merged?

JoshOrndorff:
I've merged the latest changes from `source` branch into this PR and re-cut the 1.0.0 version.

My last two remaining thoughts

* How will non-docusaurus content be versioned?
  * Rustdocs has it's own versioning system. Other items like `/recipes` should indicate which version they build against or implement their own versioning scheme
* How does this work with search results?
  * We can do [the same trick](https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/blob/source/website/siteConfig.js#L139) we did for language filtering, but we are having problems with that filtering and the rustdocs.

IMO This ca nbe merged, and we sort out the search results in the existing issues about filtering search results.

kaichaosun:
The Crowdin project exceeds the current plan, I'm figuring out what's the best option to solve it. If no, I may revert this PR.

PR #170, opened 2019-07-10
Removed inherents mention from off-chain workers doc

gautamdhameja:


gautamdhameja:
@tomusdrw from our discussion that offchain workers won't be using inherents, I made minor updates to this doc.

PR #173, opened 2019-07-11
Update "Manual Install" instructions

shawntabrizi:
https://github.com/paritytech/substrate/issues/3067

### DIFF
--- a/docs/getting-started/using-the-substrate-scripts.md
+++ b/docs/getting-started/using-the-substrate-scripts.md
@@ -1,5 +1,5 @@
 ---
-title: "Using the Substrate node and module setup scripts"
+title: "Using the Substrate Scripts"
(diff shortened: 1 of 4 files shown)


docs/getting-started/using-the-substrate-scripts.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Using the Substrate node and module setup scripts"
+title: "Using the Substrate Scripts"
```
JoshOrndorff: I like your new title better than the old one. I would like `Using the Substrate-Up Scripts` even better because

1. That's what the repo is called
2. It disambiguates substrate-up scripts from scripts/init.sh and scripts/build.sh

Up to you whether it's worth revising.

Issue #7, opened 2019-02-27
Module type


shawntabrizi:

shawntabrizi:
I am struggling a bit with this one.

Some ideas and code snippets sprawled here:
https://hackmd.io/fcQWzCFvSg2KDLm8LbJ7OA#

PR #163, opened 2019-07-06
Init runtime metadata documentation

kaichaosun:
Fix https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/34

## Background
This page targets at the blockchain developers that are not familiar with metadata in the context of Substrate. From this page, users can learn:
* What stores in the metadata, and the difference with other SRML.
* How to interact with the provided rpc method for runtime metadata.

A better reading format locates here: https://hackmd.io/@2dHpJQYgQrCqfxqXMtJTFQ/HJKGguGBS

PR #179, opened 2019-07-15
Page of Module struct

kaichaosun:
Resolve https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/7

## Why this doc
Substrate provides a convenient way to develop runtime logic. But also introduce its own DSL / context. `Module` struct is placed as a wrapper to encapsulate a module. Help users understand this struct will improve dev's confidence in using Substrate.
This page is more targeted at advanced Substrate. Part of the contents in this page can be actually put in Substrate basics after we restructure the Docs to split basics and advanced part.


docs/runtime/types/module-struct.md
```diff
@@ -0,0 +1,241 @@
...
+            let who = ensure_signed(origin)?; 
+            <Something<T>>::put(something);
+            Self::deposit_event(RawEvent::SomethingStored(something, who));
+            Ok(())
+        }
+    }
+}
+```
+
+Let's look at the code that shows the definition of `Module` struct after expanded. It has a type parameter `T` which bounds to the module's configuration `Trait`. [PhantomData](https://doc.rust-lang.org/beta/std/marker/struct.PhantomData.html) is used to indicate what type of data our `Module` struct is "tied" to, but not found in our struct.
```
shawntabrizi: I would omit the `PhantomData` stuff in your explanation and your example below. It is an implementation detail, not necessarily what you want to teach them.

docs/runtime/types/module-struct.md
```diff
@@ -0,0 +1,241 @@
...
+
+```rust
+#[cfg(feature = "std")]
+#[structural_match]
+#[rustc_copy_clone_marker]
```
shawntabrizi: Probably remove these `#[...]` items too

docs/runtime/types/module-struct.md
```diff
@@ -0,0 +1,241 @@
...
+pub struct Module<T: Trait>(::std::marker::PhantomData<(T)>);
+```
+
+> *Notes:*
+>
+> * The attribute `#[cfg(feature = "std")]` tells the Rust compiler to only compile our struct when using along with the Rust standard library.
```
shawntabrizi: If you remove it above, you can remove it here too.

docs/runtime/types/module-struct.md
```diff
@@ -0,0 +1,241 @@
...
+
+`Module` struct implements the `Callable` trait in `srml-support` by specifying the associate type to be the module's `Call` enum. Above dispatchable functions will be exposed through this enum, go to its own page for more details about [`Call` enum]((runtime/types/call-enum.md)).
+
+```rust
+impl<T: Trait> ::srml_support::dispatch::Callable for Module<T> {
+    type Call = Call<T>;
+}
+```
+
+There is also a `dispatch` function that redirect the request to the module's `Call` enum:
```
shawntabrizi: This is not so clear, but I think the meat of what we want to teach the user.

If you can expand on this, and the example below, that would be good.

docs/runtime/types/module-struct.md
```diff
@@ -0,0 +1,241 @@
...
+    pub fn dispatch<D: ::srml_support::dispatch::Dispatchable<Trait=T>>(
+        d: D,
+        origin: D::Origin,
+    ) -> ::srml_support::dispatch::Result {
+        d.dispatch(origin)
+    }
+}
+```
+
+### Implementations for Storage
```
shawntabrizi: Let's not go into this. There should be a separate page about the storage type that is generated.

docs/runtime/types/module-struct.md
```diff
@@ -0,0 +1,241 @@
...
+    type Something;
+}
+
+struct Something<T: Trait>(
+    self::sr_api_hidden_includes_decl_storage::hidden_include::storage::generator::PhantomData<
+        (T),
+    >,
+);
+
+impl<T: Trait> Store for Module<T> {
```
shawntabrizi: You should simply reference that the Module type has multple implementations, like Store, and if they want to learn more about how that implementation is generated, they can look at the Storage type.

This applies to other impl too

docs/runtime/types/module-struct.md
```diff
@@ -0,0 +1,241 @@
...
+    type Something = Something<T>;
+}
+```
+
+The `Module` struct also implements the default getter functions for storage items and the query for storage metadata. Go to [Metadata of Your Module](#metadata-of-your-module) section to give a glimpse at details. 
+
+### Deposit Event in Module
```
shawntabrizi: Same comment about Storage applies here

docs/runtime/types/module-struct.md
```diff
@@ -0,0 +1,241 @@
...
+        <system::Module<T>>::deposit_event(<T as Trait>::from(event).into());
+    }
+}
+```
+
+Refer to [Event Enum](runtime/types/event-enum.md) for more knowledge around events in Substrate.
+
+### Utilities for Block Execution
+
+There is three privileged functions that can be define in runtime modules:
```
shawntabrizi: Don't use the word "privileged" here. In Substrate, "privileged functions" are those which require ROOT origin.

I would say there are 3 "reserved" functions or "functions with reserved names"

docs/runtime/types/module-struct.md
kaichaosun: Great pickup, I found this function in implementation, but not found where is using it. I will ask other people and loop the result back here.

docs/runtime/types/module-struct.md
kaichaosun: Based on the discussion, the function is not being used and provide no value for now, remove it in this doc.

docs/runtime/types/module-struct.md
kaichaosun: * Already link details around Store in its own page, `TODO` as a mark since it does not exist yet.
* There is overlap between `Module` struct and `Event` enum, `Store` trait, put a section to document the relationship / overlap could be helpful. `Module` struct page seems a good place to do such thing.

docs/runtime/types/module-struct.md
kaichaosun: I keep the sub-sections, since it makes the structure a bit clear.

docs/runtime/types/module-struct.md
kaichaosun: Leave the PhantomData in the code example makes sure our code is syntax correct. A bit description on `PhantomData` could reduce confusion at first place. The redundancy seems valuable from this perspective.

shawntabrizi: But this phantom data description should not be found in the middle of this random page...

I dont think correct code syntax is what we are looking for here. We want to explain to them the general structures being made, not the specific nitty gritty details

docs/runtime/types/module-struct.md
```diff
@@ -0,0 +1,200 @@
...
+{}
+
+impl<T: Trait> ::srml_support::runtime_primitives::traits::OffchainWorker<T::BlockNumber>
+    for Module<T>
+{}
+```
+
+## Tuple of All The Modules
+
+Your `Module` struct then comes along with other used SRML modules to form a tuple called `AllModules` which will be tied to the `Executive` struct in `srml-executive`.
```
shawntabrizi:
```suggestion
Your `Module` struct then comes along with other used SRML modules to form a tuple called `AllModules` which will be passed to the Executive module, handling dispatch to the various modules.
```

docs/runtime/types/module-struct.md
kaichaosun: Fair point, the `PhantomData` removed.

JoshOrndorff:
I really like the `cargo expand` approach to demystifying the macros. I did wish I could see the full output of cargo expand in one place though without having to generate it myself. Not sure whether it's appropriate to link to that output here? Just a thought.

Is there a page in rustdocs about the module struct? I actually couldn't find it. Definitely this page should have a link if it exists.

What would you say are the one or two main learning objective of this post? When you are ready, I'd be happy to make a commit fixing minor English errors.

docs/runtime/types/module-struct.md
```diff
@@ -0,0 +1,200 @@
...
+pub type Executive = executive::Executive<Runtime, Block, Context, Balances, AllModules>;
+```
+
+`Executive` struct provides implementation to handle block execution, our tuple will be used during applying extrinsics. Let me stop here and leave remaining mystery for you to explore.
+
+## Metadata of Your Module
+
+`Module` struct provides query functions for metadata of module's dispatchable functions and storage, the metadata for `Event` enum is actually in its own implementation. The following json shows the metadata of our template runtime module. Go to [this page](TODO) for more details around Substrate runtime metadata.
+
+```json
```
bkchr: The Metadata is not Json. We should make this clear.

docs/runtime/types/module-struct.md
```diff
@@ -0,0 +1,200 @@
...
+        Self::deposit_event(RawEvent::SomethingStored(something, who));
+        Ok(())
+    }
+}
+```
+
+`Module` struct implements the `Callable` trait in `srml-support` by specifying the associate type to be the module's `Call` enum. Above dispatchable functions will be exposed through this enum, go to its own page for more details about [`Call` Enum](runtime/types/call-enum.md).
+
+```rust
+impl<T: Trait> ::srml_support::dispatch::Callable for Module<T> {
```
bkchr: You don't need these long paths. Just use the trait name.

kaichaosun:
@JoshOrndorff 
> I did wish I could see the full output of cargo expand in one place though without having to generate it myself.

Here is the [sample expanded code link](https://github.com/kaichaosun/substrate-real-estate-node/blob/master/runtime/expanded.rs). It's verbose and hard to read without an IDE, so it's not fit for the hub. Also if we add this link, it will redirect to a third-party site, which reduces UX and that's what we are trying to reduce. I agree with it could be valuable to have such a file. After balancing, I think we could leave this link out of the `Module` struct page, and a potential post could be a more practical way to show that.

> Is there a page in rustdocs about the module struct?

The only rustdoc I can find with `Module` struct is the [dispatch macro doc (v1.0)](https://substrate.dev/rustdocs/v1.0/srml_support/macro.decl_module.html), it's already linked in [Declaring a Module!](https://substrate.dev/docs/en/runtime/macros/decl_module) page. Compared with v1.0, the rustdocs at master branch actually provides more info, are these new docs hosted via gh pages or somewhere else?

> What would you say are the one or two main learning objective of this post? 

* What the `Module` struct looks like in standard Rust code.
* The relationship between different components (`Call` enum, `Store` trait, `Event` enum)

Feel free to commit your fix. 😄

docs/runtime/types/module-struct.md
kaichaosun: I changed the expression to emphasize what we show is in JSON format. But still don't touch metadata details. The details will be put in another [page](https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/163)

PR #187, opened 2019-07-21
Revise Session Certificate Page

joepetrowski:
Revision of Session certificate page.

cc @tomaka & @burdges


docs/overview/session-keys.md
```diff
@@ -1,24 +1,23 @@
...
+The Session certificate is a set of three key-pairs that are used by nodes and validators for network operations and signing.
 
-The default substrate node uses session keys thanks in part to the [Session module](/rustdocs/v1.0/srml_session/index.html).
+The default Substrate node implements Session keys in the [Session module](/rustdocs/v1.0/srml_session/index.html).
 
+## GRANDPA Key
 
-## BLS (Boneh-Lynn-Shacham) key
-This first key is used for consensus signatures when voting since it is desirable to be used with the GRANDPA finality gadget. BLS is preferred for voting in consensus algorithms and threshold signatures as it allows more efficient signature aggregation than when using Schnorr signatures.
+GRANDPA uses a BLS12-381 (Barreto-Lynn-Scott) key-pair for voting as it allows efficient signature aggregation with [BLS (Boneh-Lynn-Shachman) signatures](https://github.com/w3f/bls).
```
joepetrowski:
```suggestion
GRANDPA uses a BLS12-381 (Barreto-Lynn-Scott) key-pair for voting, as it allows efficient signature aggregation with [BLS (Boneh-Lynn-Shachman) signatures](https://github.com/w3f/bls).
```

docs/overview/session-keys.md
```diff
@@ -1,24 +1,23 @@
...
+A [Schnorr/Ristretto x25519 ("sr25519")](https://github.com/w3f/schnorrkel) key is used for evaluating the verifiable random function (VRF) in BABE, a block production algorithm. It was created specifically to handle Substrate and Polkadot use cases. Sr25519 implements Schnorr signatures on a [Ristretto-compressed](https://ristretto.group) ed25519 curve.
 
-## Ed25519 key (using Schnorr signatures)
-This third key is used for identifying itself to other nodes over libp2p. It is used for simpler implementations like account keys that only need to provide signature capabilities. It has a broader support ecosystem (i.e. HSMs are already available for it).
+## Network Key
 
+A Substrate node uses an ed25519 key to identify itself to other nodes over [libp2p](https://github.com/libp2p/rust-libp2p). If you run a non-validating full node, we recommend using a self-generated network key in your configuration. However, if you don't provide one, then the client will generate one for you.
```
tomaka: > If you run a non-validating full node, we recommend using a self-generated network key in your configuration. However, if you don't provide one, then the client will generate one for you.

This is relevant to the GrandPa key, I think. Not for the network key.
Users should care about their network key only if they are running a bootstrap node.

burdges: It's more correct now, but not actually correct.  

The network key should be an Ed25519 key signed by the BABE key.  Right now the grandpa key is Ed25519 but it should become a BLS key in future.

There is no "recommend" here however, either we make the node generate the transport key on start themselves, or else we do not do so.

joepetrowski: > The network key should be an Ed25519 key signed by the BABE key. Right now the grandpa key is Ed25519 but it should become a BLS key in future.

Is that Polkadot-specific? What if you are not using BABE?

Substrate takes this param, which is just a plain seed, not signed by another key. We want this page to apply to Substrate chains, not just Polkadot (within reason, let's say "Substrate for a chain that doesn't modify the Core modules").
```
--node-key <KEY>                             
            The secret key to use for libp2p networking.
```

tomaka: It's a weird situation. The code that handles the CLI options is common to all chains, but chains indeed don't necessarily use the same consensus mechanisms.

burdges: We'll need some flexible session and transport key format then.  It's not really critical that nodes create a new transport keys on startup.  We just want each node startup to register uniquely in some way that prevents the same session key being used by two nodes.

joepetrowski: I'm still not sure from this how to improve this section, except maybe to leave only the first sentence and remove:

> If you run a non-validating full node, we recommend using a self-generated network key in your configuration. However, if you don't provide one, then the client will generate one for you.

Wouldn't a validator potentially need to have several network keys? As in, the validator may be behind one or more sentry nodes. There will be one BABE key for the validator, but several sentries that present themselves as the _network_ point for a validator.

tomaka: You should remove that sentence anyway. If anything, signing the network probably won't be done manually by the user.

> Wouldn't a validator potentially need to have several network keys? As in, the validator may be behind one or more sentry nodes.

Each client has a single network key. Validators have multiple network keys, but that's because multiple clients correspond to it.

tomaka: (pinging @mxinden)

I think sentry nodes will might by passing some `--sentry-nodes` CLI option?
But you'd only ever pass one single network key.

joepetrowski: Leaving this thread open as it's interesting, but please review the update for further criticism on this PR.

mxinden: > I think sentry nodes will work by passing some --sentry-nodes CLI option, or something similar?

Yes, either a CLI option with the multi addresses of the validators or a path to a configuration file.

shawntabrizi:
@joepetrowski is this ready to review?

joepetrowski:
> @joepetrowski is this ready to review?

yes

docs/overview/session-keys.md
```diff
@@ -1,24 +1,11 @@
...
-
-## Ed25519 key (using Schnorr signatures)
-This third key is used for identifying itself to other nodes over libp2p. It is used for simpler implementations like account keys that only need to provide signature capabilities. It has a broader support ecosystem (i.e. HSMs are already available for it).
+The Session certificate is a set of keys that validators use to sign messages. `SessionKeys` is a generic, indexable type that is made concrete in the runtime. You can declare any number of keys in the Session certificate. Polkadot, for example, uses [three keys](https://wiki.polkadot.network/en/latest/polkadot/learn/keys/#session-key) in its Session certificate. Other chains could have more or fewer depending on what operations the chain expects its validators to perform.
```
bkchr: Polkadot uses 4 session keys currently.

Issue #171, opened 2019-07-10
Additional Content for Subkey Page


shawntabrizi:
Make sure you have Substrate prerequisites installed:

```
curl https://getsubstrate.io -sSf | bash -s -- --fast
```

Install the latest subkey:

```
cargo install --force --git https://github.com/paritytech/substrate subkey
```

Gav dropped this knowledge bomb and we could probably put some of it into our docs:

> an alternative, if you don't care about vanity addresses is just subkey generate
> this will provide you with a phrase that's much easier to store long-term
> e.g.
> 
> ```
> $ subkey generate
> Secret phrase `hero wheat way fury toe cheese water valve way convince ticket slogan` is account:
>   Secret seed: 0x33aa73d278cb9a5b1f76f8262d020c535959e78cfecb6b15bdd299065305808d
>   Public key (hex): 0x6ce81cff923ab6a7ac5f31dfa874c882a1b90011be6e18677f7d06cbc4e5a94f
>   Address (SS58): 5EXVyqWTPFMZgBPZmgMTy4bRx4bSkS68WbPbNJ2UejUMbQiM
> ```
> 
> if you really want to be secure, then you can up the number of words in the phrase from 12 to 24:
> 
> ```
> $ subkey generate --words 24
> Secret phrase `shoulder whale obscure culture inspire royal thumb become grape rebuild leave heavy cable cake item rubber boil use tank false skill vital notable future` is account:
>   Secret seed: 0xfeb978e544e45c59f71ed9fa13fdc25b6f5f82feeb75240cb5c62cac1dc0d2a9
>   Public key (hex): 0x7ed3549ee0941fd6e4d71182875de682d684fb3f03c2ef6269613dcf9a37bf58
>   Address (SS58): 5EvzfkVqRQmsg6RsLjQt5wEHhekthgtCSitfgPTW5juq471c
> ```
> 
> the phrase contains an in-built checksum, so there's a bit of redundancy to protect you iff it gets damaged or the handwriting is bad or whatever
> if you need multiple addresses for whatever reason, then be sure to make use of the HDKD functionality
> this way you only ever need to store away one phrase
> to use it, use subkey inspect. if you provide the phrase to it, it'll just spit out the same key as before:
> 
> ```
> $ subkey inspect "hero wheat way fury toe cheese water valve way convince ticket slogan"
> Secret phrase `hero wheat way fury toe cheese water valve way convince ticket slogan` is account:
[...]

PR #192, opened 2019-07-26
Update subkey page

joepetrowski:
Closes https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/171

Update includes:
- Dependencies
- Description of `--network`
- More password and inspect examples
- More HD key derivation explanation and examples
- Message signing and verification


shawntabrizi:
We need to duplicate this page for the versioning stuff

PR #197, opened 2019-07-29
Fix typo and clarify default value usage

jeluard:
Fix `marco` typo.
Clarify that using `config()` is not mandatory to use default storage value.

### DIFF
--- a/website/versioned_docs/version-1.0.0/runtime/initializing-storage.md
+++ b/website/versioned_docs/version-1.0.0/runtime/initializing-storage.md
@@ -10,7 +10,7 @@ In some cases storage should be initialized with a value before any user has int
 
 ## Hard-Code Default Values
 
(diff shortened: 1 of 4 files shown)


jeluard:
cc @thiolliere

website/versioned_docs/version-1.0.0/runtime/initializing-storage.md
```diff
@@ -10,18 +10,18 @@ In some cases storage should be initialized with a value before any user has int
...
     // Points ante'd to play the game
-    pub Ante get(ante) config(): u32 = 5;
+    pub Ante get(ante): u32 = 5;
 
     // Minimum raise on each turn
-    pub MinRaise get(min_raise) config(): u32 = 7;
+    pub MinRaise get(min_raise): u32 = 7;
   }
 }
 ```
```
thiolliere: on the line below (29) maybe we can change `it will revert to the default.` to `it will revert to this specified default.` to be more precise ?

shawntabrizi: fixed

PR #200, opened 2019-07-31
Remove Versioning for Now

shawntabrizi:


JoshOrndorff:
Can anyone provide some context for this decision? The latest discussion I know of is https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/118

Is the intention to target 2.0 or 1.0?

shawntabrizi:
@JoshOrndorff We did this because people were making good updates to our docs, but they were not consistently being applied to both master and v1.0.0. Furthermore, when they only got applied to master, they did not appear on the docs site since master is not the default page which shows up.

Once we have truly versioned docs, I think it makes sense to add it back. But for now, its pretty cumbersome.

PR #201, opened 2019-08-01
Update installing-substrate.md

mikewzhang:
Remove "./scripts/build.sh" because there is no such file under script folder any more. And wasm runtime is compiled while "cargo build" according to the github readme.

### DIFF
--- a/docs/getting-started/installing-substrate.md
+++ b/docs/getting-started/installing-substrate.md
@@ -121,9 +121,6 @@ Then you can build the node with
 # Update your rust toolchain
 ./scripts/init.sh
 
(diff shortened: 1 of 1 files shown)


JoshOrndorff:
There is still `scripts/build.sh` on v1.0. Are we documenting master now instead of v1.0?

shawntabrizi:
When following the instructions I updated, it does not use v1.0 since this doc attempts to mimic the `getsubstrate.io` script, which also does not use v1.0...

I agree not the best, but a detail which will solve itself sooner than making huge changes across scripts, docs, etc...

PR #202, opened 2019-08-05
Add `/presentations/` page to substrate.dev

shawntabrizi:
Just an initial design. Happy for this to be tweaked in any way.


shawntabrizi:
Maybe this should be `substrate.dev/videos/` instead? Probably a little more "memorable"...

website/data/presentations.js
```diff
@@ -0,0 +1,68 @@
...
+    summary: "Learn how the Substrate framework is built to enable forkless runtime upgrades.",
+    youtube: "https://www.youtube.com/embed/0aTnxHrV_j4",
+    link: "/presentations/Shawn Tabrizi - Substrate - Forkless Upgrades.pdf",
+    date: Date.parse("Feb 3, 2019"),
+    location: "DOT Day: Seoul, Korea",
+    weight: -1
+  }
+];
+
+function weight_then_date(a, b) {
```
kaichaosun: Function name should be in camelCase.

website/pages/en/presentations.js
```diff
@@ -0,0 +1,102 @@
...
+    const langPart = `${language ? `${language}/` : ""}`;
+    const docUrl = doc => `${baseUrl}${docsPart}${langPart}${doc}`;
+    const pageUrl = page => baseUrl + (language ? `${language}/` : "") + page;
+
+    if ((siteConfig.presentations || []).length === 0) {
+      return null;
+    }
+
+    function isFeatured(presentation) {
+      if (presentation.weight < 0) {
```
kaichaosun: Should it be bigger than 0?

shawntabrizi: The more "weight" it has, the lower to the bottom it is. If weight is negative, it "floats" to the top as featured...

Just an implementation detail

kaichaosun: This could work, just feel a bit tricky. The suggestion is either we only depend on the order in the presentation list, or only use weight: the bigger one goes to the front.

shawntabrizi:
Waiting for CA to populate initial set of presentations and to provide feedback

JoshOrndorff:
@shawntabrizi CircleCI isn't running ci on this because the topic branch is in your **fork** of the devhub repo.
image

Possible solutions:
* Open a new PR with the topic branch here in the devhub official repo
* In this one case, forget about the restrictions and merge anyway, but in the future have the topic branches live here in the official repo.

PR #208, opened 2019-08-12
Instantiable Modules

JoshOrndorff:
Questions to consider
* Should we make a module template that is _already_ instantiable? maybe as part of https://github.com/shawntabrizi/substrate-module-template
* Should I show the code snippets as diffs instead of just the new code. Diffs are more clear for those who know how to read them, but may distract newer developers.
* Should I add paragraph text before each of the snippets? It felt weird stylistically to have sections that were nothing but short code snippets, but I couldn't think of anything to write there that wasn't just filler text.
* About the video
  * Anyone know how to make it start at the start?! For me it always starts like 17 mins in?
  * Apps UI didn't properly support modules at the time of recording. It does now. I think it's worth keeping the recording, but I can understand the argument that we don't need to publish recordings of bugs.

### DIFF
--- a/docs/runtime/instantiable-module.md
+++ b/docs/runtime/instantiable-module.md
@@ -1,0 +1,140 @@
+---
+title: "Instantiable Modules"
+---
+
+Substrate runtimes are typically created by composing multiple runtime modules. Many such modules are available for re-use in the Substrate Runtime Module Library, while others may be written with a particular runtime in mind. Many such modules are intended to be re-used in a variety of runtimes. For example most blockchains have notion of a currency and may use the SRML's Balances module. But what if a single runtime wants to use the same module twice? This is exactly what instantiable modules are for.
+
+
+## What is an Instantiable Module?
+An instantiable runtime module is, in most ways, just like any other runtime module. It defines storage items, events, dispatchable calls, and a genesis configuration. But with just a few extra bits of syntax, the module can be made instantiable which means it can be deployed multiple times in the same runtime.
+
+Some use cases:
+
+* Token chain hosts two independent cryptocurrencies.
+* Marketplace track users' reputations as buyers separately from their reputations as sellers.
+* Governance has two (or more) houses which act similarly internally.
+
+The SRML's Balances and Collective modules are good examples of real-world code using this technique.
+
+## Writing an Instantiable Module
+Writing an instantiable module is almost entirely the same process as writing a plain non-instantiable module. There are just a few places where the syntax differs. If you prefer video walkthroughs, you can see a [recording](https://www.youtube.com/watch?v=XEl59hVcyI8) of making these changes.
+
+### Configuration Trait
+```rust
+pub trait Trait<I: Instance>: system::Trait {
+	// TODO: Add other types and constants required configure this module.
+
+	/// The overarching event type.
+	type Event: From<Event<Self, I>> + Into<<Self as system::Trait>::Event>;
+}
+```
+
+### Storage Declaration
+```rust
+decl_storage! {
+	trait Store for Module<T: Trait<I>, I: Instance> as TemplateModule {
+		...
+	}
+}
+```
+
+### Declaring the Module Struct
+```rust
+decl_module! {
+	/// The module declaration.
+	pub struct Module<T: Trait<I>, I: Instance> for enum Call where origin: T::Origin {
+		...
+	}
+}
+```
+### Accessing Storage
+```rust
+<Something<T, I>>::put(something);
+```
+
+<!-- In v2.0 the T is omitted for both instantiable and non-instantiable modules
+<Something<I>>::put(something); -->
+
+### Event initialization
+```rust
+fn deposit_event<T, I>() = default;
+```
+
+### Event Declaration
+```rust
+decl_event!(
+	pub enum Event<T, I> where AccountId = <T as system::Trait>::AccountId {
+		...
+	}
+}
+```
+
+## Deploying an Instantiable Module
+
+When composing a runtime, one must take care to use the proper syntax for instantiable modules. As before it is slightly different.
+
+### Implementing the Configuration Trait
+When implementing the module's configuration trait, you need to be explicit which instance you are configuring, because in general the instances need not have identical configurations. In this snippet we've named the instance `Instance1`.
+```rust
+impl template::Trait<template::Instance1> for Runtime {
+	type Event = Event;
+}
+```
+
+### In the `construct_runtime!` Macro
+Finally, in the `construct_runtime!` macro, we give a more human-friendly name to each instance of the module. Here I've called `Instance1` `FirstTemplate`.
+```rust
+FirstTemplate: template::<Instance1>::{Module, Call, Storage, Event<T>, Config},
+```
+
+The default Substrate node does a similar thing with its council and technical committee.
+```rust
+Council: collective::<Instance1>::{Module, Call, Storage, Origin<T>, Event<T>, Config<T>},
+TechnicalCommittee: collective::<Instance2>::{Module, Call, Storage, Origin<T>, Event<T>, Config<T>}
+```
+
+## Default Instance
+One drawback of instantiable modules, as we've presented them so far is that they require the deployer to use the more elaborate syntax even if they only desire a single instance of the module. To alleviate this inconvenience, Substrate provides a feature known as DefaultInstance. This allows runtime developers to deploy an instantiable module exactly as they would if it were not instantiable provided they **only use a single instance**.
+
+To make your instantiable module support DefaultInstance, you must specify it in three places.
+
+```rust
+pub trait Trait<I=DefaultInstance>: system::Trait {
+```
+
+```rust
+decl_storage! {
+  trait Store for Module<T: Trait<I>, I: Instance=DefaultInstance> as TemplateModule {
+    ...
+  }
+}
+```
+
+```rust
+decl_event!(
+	pub enum Event<T, I=DefaultInstance> where ... {
+    ...
+  }
+}
+```
+
+Having made these three changes, a developer who uses your module doesn't need to know or care that your module is instantable. They can deploy it just as they would any other module.
+
+## Genesis Configuration
+Some modules require a genesis configuration to be specified. This process is identical for instantiable and non-instantiable modules, but it is worth mentioning that the configuration is named after the human-friendly name given in `construct_runtime!`, and is different for each instance. That is to say that the configuration is **not** named after the module itself. Let's look to the default Substrate node's use of the Collective module as an example.
+
+In its `chain_spec.rs` file we see
+```rust
+GenesisConfig {
+	...
+	collective_Instance1: Some(CouncilConfig {
+		members: vec![],
+		phantom: Default::default(),
+	}),
+	collective_Instance2: Some(TechnicalCommitteeConfig {
+		members: vec![],
+		phantom: Default::default(),
+	}),
+	...
+}
+```
--- a/docs/runtime/instantiable-module.md
+++ b/docs/runtime/instantiable-module.md
@@ -6,7 +6,7 @@ Substrate runtimes are typically created by composing multiple runtime modules. 
 
 
 ## What is an Instantiable Module?
-An instantiable runtime module is, in most ways, just like any other runtime module. It defines storage items, events, dispatchable calls, and a genesis configuration. But with just a few extra bits of syntax, the module can be made instantiable which means it can be deployed multiple times in the same runtime.
+An instantiable runtime module is, in most ways, just like any other runtime module. It defines storage items, events, dispatchable calls, and a genesis configuration. However, instantiable modules can have multiple instances included in a single runtime. Each instance of the module has its own independent storage, and extrinsics must specify which instance of the module they are intended for.
 
 Some use cases:
 
@@ -14,10 +14,14 @@ Some use cases:
 * Marketplace track users' reputations as buyers separately from their reputations as sellers.
 * Governance has two (or more) houses which act similarly internally.
 
-The SRML's Balances and Collective modules are good examples of real-world code using this technique.
+The SRML's Balances and Collective modules are good examples of real-world code using this technique. The default Substrate node has two instances of the Collectives module that make up its Council and Technical Committee. Each collective has its own storage, events, and configuration.
 
 ## Writing an Instantiable Module
 Writing an instantiable module is almost entirely the same process as writing a plain non-instantiable module. There are just a few places where the syntax differs. If you prefer video walkthroughs, you can see a [recording](https://www.youtube.com/watch?v=XEl59hVcyI8) of making these changes.
+
+> You must call `decl_storage!`
+>
+> Instantiable modules _must_ call the `decl_storage!` macro so that the `Instance` type is created.
 
 ### Configuration Trait
 ```rust
@@ -69,20 +73,20 @@ decl_event!(
 }
 ```
 
-## Deploying an Instantiable Module
+## Installing a Module Instsance in a Runtime
 
-When composing a runtime, one must take care to use the proper syntax for instantiable modules. As before it is slightly different.
+The syntax for including an instance of an instantiable module in a runtime is slightly different than for a regular module. The only exception is for modules that use the [Default Instance](#defualt-instance) feature described below.
 
-### Implementing the Configuration Trait
-When implementing the module's configuration trait, you need to be explicit which instance you are configuring, because in general the instances need not have identical configurations. In this snippet we've named the instance `Instance1`.
+### Implementing Configuration Traits
+Each instance needs to be configured individually. You must implement the module's configuration trait once for each instance you have installed, and you must specify which instance the configuration is for. In the following snippet we are configuring `Instance1`.
 ```rust
 impl template::Trait<template::Instance1> for Runtime {
 	type Event = Event;
 }
 ```
 
-### In the `construct_runtime!` Macro
-Finally, in the `construct_runtime!` macro, we give a more human-friendly name to each instance of the module. Here I've called `Instance1` `FirstTemplate`.
+### Using the `construct_runtime!` Macro
+The final step of installing the module instance in your runtime is updating the `construct_runtime!` macro. You may give each instance a meaningful name. Here I've called `Instance1` `FirstTemplate`.
 ```rust
 FirstTemplate: template::<Instance1>::{Module, Call, Storage, Event<T>, Config},
 ```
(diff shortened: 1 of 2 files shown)


shawntabrizi:
> Anyone know how to make it start at the start?! 

This is a youtube caching thing. For me, and anyone who has not watched the video, it starts from the beginning. Otherwise you can add a `&start=0` if you want to force it, but probably not recommended.

I am a big fan of embedding the video at the bottom of the page, and then have the link pull the user to the bottom?

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
+---
+title: "Instantiable Modules"
+---
+
+Substrate runtimes are typically created by composing multiple runtime modules. Many such modules are available for re-use in the Substrate Runtime Module Library, while others may be written with a particular runtime in mind. Many such modules are intended to be re-used in a variety of runtimes. For example most blockchains have notion of a currency and may use the SRML's Balances module. But what if a single runtime wants to use the same module twice? This is exactly what instantiable modules are for.
```
shawntabrizi: I am not personally a fan of how this is written. This is part of the docs for the runtime, we should have already explained what a runtime is, the fact that modules can be written etc...

Something much more succinct I think would be better here:

"Substrate allows you to use the same runtime module multiple times in the same Substrate runtime through the instantiable modules feature" or something like that.

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
...
+Some use cases:
+
+* Token chain hosts two independent cryptocurrencies.
+* Marketplace track users' reputations as buyers separately from their reputations as sellers.
+* Governance has two (or more) houses which act similarly internally.
+
+The SRML's Balances and Collective modules are good examples of real-world code using this technique.
+
+## Writing an Instantiable Module
+Writing an instantiable module is almost entirely the same process as writing a plain non-instantiable module. There are just a few places where the syntax differs. If you prefer video walkthroughs, you can see a [recording](https://www.youtube.com/watch?v=XEl59hVcyI8) of making these changes.
```
shawntabrizi: If you think the video should stay on the post, I would embed it and put it in a section at the bottom named `# Video` and then link to it.

shawntabrizi: We should make this a common pattern, so people can go to pages and just add #video at the end or whatever

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
+---
+title: "Instantiable Modules"
+---
+
+Substrate runtimes are typically created by composing multiple runtime modules. Many such modules are available for re-use in the Substrate Runtime Module Library, while others may be written with a particular runtime in mind. Many such modules are intended to be re-used in a variety of runtimes. For example most blockchains have notion of a currency and may use the SRML's Balances module. But what if a single runtime wants to use the same module twice? This is exactly what instantiable modules are for.
+
+
+## What is an Instantiable Module?
+An instantiable runtime module is, in most ways, just like any other runtime module. It defines storage items, events, dispatchable calls, and a genesis configuration. But with just a few extra bits of syntax, the module can be made instantiable which means it can be deployed multiple times in the same runtime.
```
shawntabrizi: "But with just a few extra bits of syntax"

Really not a fan of this for a docs page.

Also you did not really explain what an instantiable module is technically.

You should say comment about:
* Multiple module instantiations
* Multiple dispatch routing
* Multiple unique runtime storage

and the other technical aspects of what really an instantiable module is

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
...
+* Token chain hosts two independent cryptocurrencies.
+* Marketplace track users' reputations as buyers separately from their reputations as sellers.
+* Governance has two (or more) houses which act similarly internally.
+
+The SRML's Balances and Collective modules are good examples of real-world code using this technique.
+
+## Writing an Instantiable Module
+Writing an instantiable module is almost entirely the same process as writing a plain non-instantiable module. There are just a few places where the syntax differs. If you prefer video walkthroughs, you can see a [recording](https://www.youtube.com/watch?v=XEl59hVcyI8) of making these changes.
+
+### Configuration Trait
```
shawntabrizi: Why is there just a block of text with no description for all of these sections?

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
...
+### Event Declaration
+```rust
+decl_event!(
+	pub enum Event<T, I> where AccountId = <T as system::Trait>::AccountId {
+		...
+	}
+}
+```
+
+## Deploying an Instantiable Module
```
shawntabrizi: "Adding an Instantiable Module to your Runtime" or something.

Deploying seems like the wrong word in this context

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
...
+
+When composing a runtime, one must take care to use the proper syntax for instantiable modules. As before it is slightly different.
```
shawntabrizi: Lets use "you" rather than "one".

```suggestion
When composing your runtime, you must take care to use the proper syntax needed for instantiable modules.
```

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Implementing the Configuration Trait
+When implementing the module's configuration trait, you need to be explicit which instance you are configuring, because in general the instances need not have identical configurations. In this snippet we've named the instance `Instance1`.
```
shawntabrizi: Would say it more like:

"Each instance of a module needs to be individually configured. When implementing the module's configuration trait, you must specify the instance you are configuring. In the example below, we configure the instance named `Instance`."

shawntabrizi: I think the general feeling I get is that you are making suggestions to me, rather than telling me what should be done. If it is a fact, use words like "must" and "need"

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
...
+```rust
+impl template::Trait<template::Instance1> for Runtime {
+	type Event = Event;
+}
+```
+
+### In the `construct_runtime!` Macro
+Finally, in the `construct_runtime!` macro, we give a more human-friendly name to each instance of the module. Here I've called `Instance1` `FirstTemplate`.
```
shawntabrizi: "we give a more human-friendly name to each instance of the module"

This is absolutely not required when writing instantiate modules. You can do it in the examples for clarity, but I would omit this kind of text from the docs.

What you are doing when you add it to `construct_runtime` is that you are _actually_ adding the module to your runtime. If the module is not put here, it will not be "constructed" into the runtime.

This is why this step is important and what you should be telling the user.

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Implementing the Configuration Trait
+When implementing the module's configuration trait, you need to be explicit which instance you are configuring, because in general the instances need not have identical configurations. In this snippet we've named the instance `Instance1`.
+```rust
+impl template::Trait<template::Instance1> for Runtime {
+	type Event = Event;
+}
+```
+
+### In the `construct_runtime!` Macro
```
shawntabrizi: Weird title here. "Add Your Module To Your Runtime" or something

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
...
+Finally, in the `construct_runtime!` macro, we give a more human-friendly name to each instance of the module. Here I've called `Instance1` `FirstTemplate`.
+```rust
+FirstTemplate: template::<Instance1>::{Module, Call, Storage, Event<T>, Config},
+```
+
+The default Substrate node does a similar thing with its council and technical committee.
```
shawntabrizi: I think we should bring this example up to the top when you introduce "What is instantible modules"

I think it would be good to go more into detail about how the Substrate node uses this feature. We have two councils, they have their own storage, own functions, etc...

Then I would omit it here.

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
...
+```rust
+Council: collective::<Instance1>::{Module, Call, Storage, Origin<T>, Event<T>, Config<T>},
+TechnicalCommittee: collective::<Instance2>::{Module, Call, Storage, Origin<T>, Event<T>, Config<T>}
+```
+
+## Default Instance
+One drawback of instantiable modules, as we've presented them so far is that they require the deployer to use the more elaborate syntax even if they only desire a single instance of the module. To alleviate this inconvenience, Substrate provides a feature known as DefaultInstance. This allows runtime developers to deploy an instantiable module exactly as they would if it were not instantiable provided they **only use a single instance**.
```
shawntabrizi: Again, for docs I think this can be written way more succinctly.

"You can use the `DefaultInstance` trait(?) to enable only a single instance of an instantiable module."

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
...
+
+To make your instantiable module support DefaultInstance, you must specify it in three places.
```
shawntabrizi: Why would they not want to support this? Should it just be part of the main instructions for making an instantiable module?

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
...
+	pub enum Event<T, I=DefaultInstance> where ... {
+    ...
+  }
+}
+```
+
+Having made these three changes, a developer who uses your module doesn't need to know or care that your module is instantable. They can deploy it just as they would any other module.
+
+## Genesis Configuration
+Some modules require a genesis configuration to be specified. This process is identical for instantiable and non-instantiable modules, but it is worth mentioning that the configuration is named after the human-friendly name given in `construct_runtime!`, and is different for each instance. That is to say that the configuration is **not** named after the module itself. Let's look to the default Substrate node's use of the Collective module as an example.
```
shawntabrizi: "the configuration is **not** named after the module itself."

This is really not helpful. Just say what you want to say:

"When defining the genesis config for an instantiable module, you must append the "Instance Name" to the name of the module.

For example, the `TechnicalComittee` and `Council` in the main Substrate node both use the `collective` module:

```rust
Council: collective::<Instance1>::{Module, Call, Storage, Origin<T>, Event<T>, Config<T>},
TechnicalCommittee: collective::<Instance2>::{Module, Call, Storage, Origin<T>, Event<T>, Config<T>}
```

You would then configure these modules like so:
...

shawntabrizi:
Have you covered all the features of Instantiable modules?

@thiolliere can you confirm?

In general, I think we need to tune the style of these kinds of docs pages.

Certainly friendly and readable, but less "suggestive" and no "opinions".

It should feel like:

"If you do X, then Y."

"You (must/may/must not) do X."

Also, our users are technical, and so I feel they will want a more technical understanding of how this feature works. How is it different from creating two modules? What is happening in the background?

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
...
+	}
+}
+```
+### Accessing Storage
+```rust
+<Something<T, I>>::put(something);
+```
+
+<!-- In v2.0 the T is omitted for both instantiable and non-instantiable modules
+<Something<I>>::put(something); -->
```
thiolliere: This depends on if the type is generic over T or not I think no ?

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Event initialization
+```rust
+fn deposit_event<T, I>() = default;
```
thiolliere: isn't it possible to declase an event that is not generic over T ?

thiolliere:
Also an instantiable module must call decl_storage so the trait Instance and also maybe DefaultInstance type are created https://github.com/paritytech/substrate/pull/3451/files

docs/runtime/instantiable-module.md
thiolliere: note: it kind of accept not generic event at this stage but then construct_runtime will complains because it needs Event to be generic over instance at least. (I think we should warn or compile-error in case someone write a non-generic deposit event here.)

docs/runtime/instantiable-module.md
JoshOrndorff: > It felt weird stylistically to have sections that were nothing but short code snippets, but I couldn't think of anything to write there that wasn't just filler text.

docs/runtime/instantiable-module.md
JoshOrndorff: Grammar girl agrees with you, so I'll avoid "one" https://www.quickanddirtytips.com/education/grammar/one-versus-you :)

I just wrote around it instead of using "you" though.

docs/runtime/instantiable-module.md
JoshOrndorff: Agree with the first comment. Not sure about the second. I used exactly those words.

docs/runtime/instantiable-module.md
JoshOrndorff: > This is absolutely not required when writing instantiate modules.

Yes, I agree, but

> If the module is not put here, it will not be "constructed" into the runtime.

gives the incorrect impression that removing the module from `construct_runtime!` is sufficient to remove it from the runtime. In fact making that change alone simply causes the runtime not to compile which is why I didn't use that phrasing.

I've edited in a way that I think addresses both concerns.

docs/runtime/instantiable-module.md
JoshOrndorff: I had already mentioned that use case
> Governance has two (or more) houses which act similarly internally.

I've now added a sentence stating that the default Substrate node has two Collectives.

docs/runtime/instantiable-module.md
JoshOrndorff: > You can use the DefaultInstance trait(?) to enable only a single instance of an instantiable module.

That misses the point. One can include only a single instance regardless of whether DefaultInstance is used. The point of DefaultInstance is that the runtime designer doesn't need to know or care that that the module was instantiable or use the longer syntax.

docs/runtime/instantiable-module.md
JoshOrndorff: > Why would they not want to support this?

I wondered that too. I can't think of a use case.

> Should it just be part of the main instructions for making an instantiable module?

That would make the article shorter, but we would lose the opportunity to explain exactly what DefaultInstance actually does. Personally I prefer it the way it is. One topic at a time.

docs/runtime/instantiable-module.md
JoshOrndorff: I would be happy to know more about this. I based my comment on the template module
https://github.com/paritytech/substrate/blob/master/node-template/runtime/src/template.rs#L49

docs/runtime/instantiable-module.md
JoshOrndorff: Good to know. And I agree that warning/error would be good. But the doc should not change right? No need to point users toward something that will cause them trouble.

JoshOrndorff:
I've applied the suggested changes that I either:
* agree with
* am indifferent about

I've left the others unchanged.

docs/runtime/instantiable-module.md
thiolliere: I linked a diff showing an example.
[example.txt](https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/files/3534978/example.txt)

docs/runtime/instantiable-module.md
4meta5: I also think the first sentence should clearly communicate the value add of instantiable modules. Something like 

"Instantiable Modules enable the use of the same module in different ways within a runtime. Polkadot's governance uses this feature to incorporate multiple councils for screening proposals. Because each council uses the same module, but chooses different configurations, the instancing pattern described below is especially useful."

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,144 @@
+---
+title: "Instantiable Modules"
+---
+
+Substrate runtimes are typically created by composing multiple runtime modules. Many such modules are available for re-use in the Substrate Runtime Module Library, while others may be written with a particular runtime in mind. Many such modules are intended to be re-used in a variety of runtimes. For example most blockchains have notion of a currency and may use the SRML's Balances module. But what if a single runtime wants to use the same module twice? This is exactly what instantiable modules are for.
+
+
+## What is an Instantiable Module?
+An instantiable runtime module is, in most ways, just like any other runtime module. It defines storage items, events, dispatchable calls, and a genesis configuration. However, instantiable modules can have multiple instances included in a single runtime. Each instance of the module has its own independent storage, and extrinsics must specify which instance of the module they are intended for.
```
4meta5: The first sentence seems unnecessary. The second sentence could open like this

"Unlike other runtime modules, instantiable modules enable multiple instances of the same module logic within a single runtime."

and the last sentence works as is..

docs/runtime/instantiable-module.md
4meta5: I think not having any description here is fine because there are comments and the code is self-explanatory. If the reader doesn't know what the `Trait` object is, they should start somewhere else anyway.

docs/runtime/instantiable-module.md
```diff
@@ -0,0 +1,144 @@
...
+	}
+}
+```
+
+## Installing a Module Instsance in a Runtime
+
+The syntax for including an instance of an instantiable module in a runtime is slightly different than for a regular module. The only exception is for modules that use the [Default Instance](#defualt-instance) feature described below.
+
+### Implementing Configuration Traits
+Each instance needs to be configured individually. You must implement the module's configuration trait once for each instance you have installed, and you must specify which instance the configuration is for. In the following snippet we are configuring `Instance1`.
```
4meta5: "individually" and "installed" may not be the best words for describing this, but they may be sufficient. Here's another version -- ignore if you feel like it.

```suggestion
Each instance must be configured separately. Configuration consists of implementing the specific instance's trait once for each instance. The following snippet shows a configuration for `Instance1`.
```

docs/runtime/instantiable-module.md
4meta5: I agree with @shawntabrizi and think we should put the code snippet at the top to make it clear what implementing this feature looks like and link to where it's used. Showing the governance configuration at the top also gives the reader a good reference when they're looking at the examples provided in this doc and trying to connect that to how they might use the feature in their codebase.

docs/runtime/instantiable-module.md
4meta5: I feel like this is a tangent for someone that has implemented an instantiable module but doesn't want to make it instantiable. It seems unnecessary to cover here; maybe an italicized note somewhere with a link to documentation would suffice like

```
Note: Instantiable modules that contain only a single instance can implement [`DefaultInstance`](INSERT_LINK) for configuration when the runtime is not taking advantage of the module's instancing capabilities.
```

docs/runtime/instantiable-module.md
4meta5: I agree and think it is easier to read if you just cut the second to last sentence.

docs/runtime/instantiable-module.md
shawntabrizi: Let's not start dropping the "P" word unless we are writing something specific about Polkadot.

docs/runtime/instantiable-module.md
JoshOrndorff: I included this specifically because I got stuck on it while learning. I thought I would use `CollectiveConfig` twice as opposed to `CouncilConfig` and `TechnicalCommitteeConfig`.

However I appreciate the peer review and have taken the suggestion.

docs/runtime/instantiable-module.md
4meta5: I like using the "P" word. It's the best (only) reference implementation that demonstrates how this feature is actually used in practice. It provides context and, without context, the application of this feature is ambiguous and not clear.

PR #217, opened 2019-08-28
Joshy instantiable followup

JoshOrndorff:
This is a follow-up to https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/208 which I accidentally merged before pushing the changes from the second round of feedback.

### DIFF
--- a/docs/runtime/instantiable-module.md
+++ b/docs/runtime/instantiable-module.md
@@ -2,11 +2,7 @@
 title: "Instantiable Modules"
 ---
 
-Substrate runtimes are typically created by composing multiple runtime modules. Many such modules are available for re-use in the Substrate Runtime Module Library, while others may be written with a particular runtime in mind. Many such modules are intended to be re-used in a variety of runtimes. For example most blockchains have notion of a currency and may use the SRML's Balances module. But what if a single runtime wants to use the same module twice? This is exactly what instantiable modules are for.
-
-
-## What is an Instantiable Module?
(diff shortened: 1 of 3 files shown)


docs/runtime/instantiable-module.md
```diff
@@ -2,11 +2,7 @@
...
-
-
-## What is an Instantiable Module?
-An instantiable runtime module is, in most ways, just like any other runtime module. It defines storage items, events, dispatchable calls, and a genesis configuration. However, instantiable modules can have multiple instances included in a single runtime. Each instance of the module has its own independent storage, and extrinsics must specify which instance of the module they are intended for.
+Unlike other runtime modules, instantiable modules enable multiple instances of the same module logic within a single runtime. Each instance of the module has its own independent storage, and extrinsics must specify which instance of the module they are intended for.
```
shawntabrizi: > Unlike other runtime modules, instantiable modules

You make it sound like instantiable modules are different than runtime modules...

I would phrase this to say more like "Runtime modules have the ability to blah blah with the instantiable module feature..." but that is just me.

shawntabrizi:
```suggestion
On this page we will teach you how to create an "Instantiable Module", which allows you to enable multiple instances of the same module logic within a single runtime. Each instance of the module has its own independent storage, and extrinsics must specify which instance of the module they are intended for.
```

4meta5: agree!

Issue #222, opened 2019-08-29
Add Branch Protections and Checks


shawntabrizi:
* Check Source Branch builds before new commits are merged
* Make source branch only allow pull requests, no direct changes to master
* Make so that `master` branch can only be updated by our deployment bot

* other?

JoshOrndorff:
The second item means that nobody can push directly to `source` but instead we have to make PRs right?

If so, I think these are all a good ideas.

But I don't have permissions to set that up.

shawntabrizi:
@JoshOrndorff try again now?

JoshOrndorff:
@shawntabrizi Yep, I have access now, Thanks.



[ ] Check Source Branch builds before new commits are merged

[x] Make source branch only allow pull requests, no direct changes to master

[x] Make so that master branch can only be updated by our deployment bot

shawntabrizi:
The build checks are not actually completing

JoshOrndorff:
The issue was that branches in forks were not triggering the build. @shawntabrizi has confirmed it's resolved now.

PR #225, opened 2019-08-30
Update CI config to check each build

JoshOrndorff:
In https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/222 we agreed that no PRs should be merged into `source` unless the build passes in CI. This PR enables a job which checks the build. Once this goes in, I can require that the build pass in github settings and we can close the issue.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ jobs:
             crowdin --config crowdin.yaml download -b master
             # build and publish website
             cd website && USE_SSH=true GIT_USER=git yarn run publish-gh-pages
-  
+
   check-build:
     docker:
       # specify the version you desire here
@@ -67,3 +67,6 @@ workflows:
           filters: *filter-only-source
     jobs:
       - deploy-website
+  check_build:
+    jobs:
+      - check-build


.circleci/config.yml
```diff
@@ -67,3 +67,6 @@ workflows:
           filters: *filter-only-source
     jobs:
       - deploy-website
+  check_build:
+    jobs:
+      - check-build
```
kaichaosun: It seems every commit in this repo will trigger the build, but PR from 3rd party will not trigger this workflow, is this correct?

Given Circle CI does not [support build](https://ideas.circleci.com/ideas/CCI-I-316) directly on PR, I agree with the current solution.

shawntabrizi: I will take this as an approval

JoshOrndorff: @kaichaosun I think this is good enough for now, but you raise a good point.

https://discuss.circleci.com/t/how-to-trigger-pr-pipelines-in-circleci-for-bitbucket/31715 gave me the idea that we could use github's webhooks to trigger ci builds when PRs are open. Maybe that's worth exploring if the current solution becomes a bottleneck.

Issue #227, opened 2019-09-02
Add Google Analytics with Klaro Consent


shawntabrizi:
We want to add google analytics to Substrate Developer Hub, but we also want to ensure that we are GDPR compliant.

Fortunately, we can use Klaro to do these all without a server.

Here is how it is done for the Kusama Network site:
https://github.com/kusamanetwork/kusama.network/commit/e6e3a7f3d9772a222bf968d8e201fe695c217ff6

We need to use our own fork of the repo to use local storage rather than cookies:
https://github.com/KIProtect/klaro/pull/84

Should also look to add this across all our resources, maybe we can make a generalized plugin for this purpose so we just add one line or something.

---
TODOs
- [x] Substrate developer hub
- [x] Substrate collectables workshop
- [x] Substrate contracts workshop
- [x] Recipes
- [x] Rust docs
- [ ] Support chinese translation, [upstream issue](https://github.com/KIProtect/klaro/issues/1)

shawntabrizi:
We can probably create a centralized plugin that can be reused everywhere, and thus only a single consent is needed across all the substrate.dev sub-sites.

JoshOrndorff:
I'm fine with this implementation if we choose to use analytics. I just want to at least consider not using analytics for all the reasons that Stallman talked about at web3Summit.

kaichaosun:
It seems we have more options:
* Use consolidated GA account
* [Matomo](https://matomo.org/)
* [Snowplow](https://github.com/snowplow/snowplow)

I will use a standalone google analytics account for initial implementation.

shawntabrizi:
@kaichaosun We already have a GA account.

Please use these settings:

```
<!-- Global site tag (gtag.js) - Google Analytics -->
<script async src="https://www.googletagmanager.com/gtag/js?id=UA-145158313-2"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());

  gtag('config', 'UA-145158313-2');
</script>
```

shawntabrizi:
You can also see how they integrated Klaro into Parity.io:

https://github.com/paritytech/paritytech.github.io/commits/master

shawntabrizi:
@kaichaosun We still want to add these same components to other projects in the Substrate Develop Hub, for example the Kitties tutorial.

Is there some code we can generally copy and paste to make that happen?

[...]

PR #230, opened 2019-09-06
Integrate Klaro and Google Analytics

kaichaosun:
Fix: https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/227
TODO: 
- [x] Link to privacy policy page.
- [x] Fix the klara in chinese pages.


website/core/Footer.js
```diff
@@ -103,7 +103,21 @@ class Footer extends React.Component {
           </div>
         </section>
 
-        <section className="copyright">{this.props.config.copyright}</section>
+        <section className="row">
+          <div className="copyright">{this.props.config.copyright}</div>
+          <a className="item" href="https://www.parity.io/privacy/">Privacy Policy</a>
+          <a className="item" href="#" id="cookie-settings">Cookie Settings</a>
+          <script
```
kaichaosun: Why ugly script here? => [reason](https://github.com/facebook/Docusaurus/issues/584)

website/siteConfig.js
```diff
@@ -77,7 +77,18 @@ const siteConfig = {
...
-  scripts: ['https://buttons.github.io/buttons.js'],
+  scripts: [
+    'https://buttons.github.io/buttons.js',
+    '/js/load.js',
+    {
+      src: '/js/config.js',
+      defer: true,
+    },
+    {
+      src: '/js/klaro.min.js',
```
JoshOrndorff: Good idea to use the minified version. I see this PR includes both the minified and original versions of the file though. Do we need both?

website/static/css/custom.css
```diff
@@ -72,6 +72,31 @@
...
 }
 
+/* Fix the klaro */
+.klaro .cookie-modal .cm-modal .hide {
+    top: 20px !important;
+    display: inline-block;
+}
+
+.cm-toggle-all {
+    display: none !important;
```
JoshOrndorff: What is `.cm-toggle-all`? I didn't notice anything changine when I removed this rule. And is `display: none;` the best way to disable it? Maybe it is, but red flags go off when I see `!important`. (Same applies to the `top` rule above.

I found this article helpful about `!important` https://css-tricks.com/when-using-important-is-the-right-choice/

ButI actually didn't notice these css rules doing anything.

website/static/js/config.js
```diff
@@ -0,0 +1,153 @@
...
+      },
+      purposes: {
+        analytics: 'Besucher-Statistiken',
+        security: 'Sicherheit',
+      },
+    },
+    en: {
+      consentModal: {
+        description:
+          'We use necessary cookies to make our site work. We\'d also like to set optional tracking mechanisms to help us improve it by collecting and reporting information on how you use it. Here you can customise the information we collect about you by giving or revoking your consent to our use of optional tracking mechanisms.',
```
JoshOrndorff: What "necessary cookies" do we use? Isn't this just about google analytics which we're allowing the user to opt out of?

website/static/js/config.js
```diff
@@ -0,0 +1,153 @@
...
+      },
+      cloudflare: {
+        description: 'Protection against DDoS attacks',
+      },
+      purposes: {
+        analytics: 'analytics and improvement of our sites',
+        security: 'security',
+      },
+    },
+    tr: {
```
JoshOrndorff: It seems our second most popular language is Chinese. Should we have that translation here? Also how does Klaro detect the language to use. Will it automatically show the consent box in the correct language if the user selects a different language for devhub content?

website/static/js/config.js
```diff
@@ -0,0 +1,153 @@
...
+      consentModal: {
+        description:
+          'We use necessary cookies to make our site work. We\'d also like to set optional tracking mechanisms to help us improve it by collecting and reporting information on how you use it. Here you can customise the information we collect about you by giving or revoking your consent to our use of optional tracking mechanisms.',
+      },
+      googleAnalytics: {
+        description: 'Collection of information about how visitors use our website',
+      },
+      cloudflare: {
+        description: 'Protection against DDoS attacks',
+      },
```
JoshOrndorff: Are we actually using cloudflare for anything, or is this from the template? Does it hurt to leave it in?

website/static/js/config.js
shawntabrizi:
```suggestion
          'We use necessary cookies to make our site work. We\'d also like to set optional tracking mechanisms to help us improve it by collecting and reporting information on how you use it. Here you can customize the information we collect about you by giving or revoking your consent to our use of optional tracking mechanisms.',
```

website/static/js/config.js
shawntabrizi: I think this is from the Parity.io site. I don't think we use cloudflare, but maybe we should. Need to talk to devops, for now we can probably remove it.

shawntabrizi: I don't see it show up in the UI though, so...
image

website/siteConfig.js
shawntabrizi: I agree, the original version is not very readable either, we could just drop it

website/static/css/custom.css
shawntabrizi: `cm-toggle-all` is the element which allows you to toggle all "klaro apps" at once:

image


We only have one choice, so I agree, it makes sense to hide it, unless there is a specific configuration we can set rather than hiding it with CSS

website/static/js/config.js
shawntabrizi: +1, but I don't think should be a blocker to be merged. If we can address it in a follow up PR, that is fine too

shawntabrizi:
Possible bug?

On `localhost` if I look at the cookie settings, and turn google analytics on, and then I refresh the page, and look at cookie settings again, it is now turned off.

I also do not see the prompt....

website/static/js/config.js
shawntabrizi: Yeah, without cloudflare, I think there are no "necessary" cookies.

This means we should probably remove the first sentence, and change the two buttons on the original klaro prompt to represent you either allow GA or you allow nothing.

website/static/js/config.js
kaichaosun: Klaro doesn't support Chinese yet, https://github.com/KIProtect/klaro/blob/master/src/translations/index.js

website/static/js/config.js
kaichaosun: The current text in buttons is consistent with parity.io and kusama.network (it also has GA only).
The button can be renamed by text changes in the klaro.min.js. Kirill @kirushik may be the person can help the decision on it.

kaichaosun:
> On localhost if I look at the cookie settings, and turn google analytics on, and then I refresh the page, and look at cookie settings again, it is now turned off.

I haven't reproduced this scenario, we can work together to figure it out.

website/static/js/config.js
kaichaosun: Klaro fixed in Chinese page after update it to the latest.
Later PR will add Chinese support.

website/static/js/config.js
shawntabrizi: Yeah, let's have Kirill decide what to put

kirushik: I've seen we're thinking about integrating substrate playground into substrate.dev.
Since it will feature logins, it would have to rely on some cookies mechanism, right?
So we either keep this in place for now, or we would have to rebuild klaro while integrating playground in our site.
Keeping it seems to be an easier solution, both because it requires less effort, and because there's a lower chance to forget it and go to production with inaccurate disclaimers.

kaichaosun:
@kirushik The "toggle all" doesn't work as expected. Since we want the cloudflare to be always required, but "toggle all" can disable it in storage but not make it gray in the UI. We better hide it so that it won't confuse the user IMO.

kirushik:
@kaichaosun OK, I've just discovered that we're doing the same CSS trick for hiding the "toggle all" option on parity.io as well — so it's OK to use it here as well.
Other than that — nice job, let's merge it :shipit: !

shawntabrizi:
WOO!

PR #236, opened 2019-09-08
Add Kusama to the list > Who is building on Substrate?

iammelea:
Add Kusama project to the list of projects building on substrate
https://substrate.dev/en/users.html


shawntabrizi:
@iammelea you need to include an image right?

iammelea:
> @iammelea you need to include an image right?

is included

PR #243, opened 2019-09-16
Make image square

shawntabrizi:


shawntabrizi:
Maybe there is a CSS solution, in general, but this is easiest

PR #245, opened 2019-09-17
Fix google analytics

kaichaosun:
Many thanks for the help from @axelchalon


kaichaosun:
Test in local, now it can collect data.

shawntabrizi:
Yeah it looks good, and seems to work from my initial test.

PR #246, opened 2019-09-17
Add new Polkadot JS API Tutorial

shawntabrizi:
This adds Jaco's tutorial to the list of "front-end" tutorials


shawntabrizi:
Good point. It is kind of somewhere in between right? It feels a lot like a guide/codelab to me (although there is not much explicit actions for the user).

You are probably right that it shoudn't be in the tutorial section.

I think we will have to point externally to the Polkadot JS docs pretty much no matter what. I think the same will be true for some of the Polkadot Network research docs.

I see no reason to put limits on how we link outward. Obviously it shouldn't link outward in our table of contents. It could also make sense that we have external links open a new tab?

PR #250, opened 2019-09-19
Proposal for V2.0 Documentation

shawntabrizi:
This PR is draft PR which suggests a huge overhaul to the existing Substrate Documentation, following the layout of our new Contribution Guidelines. #247 

At a high level, we want to make sure our documentation provides:
* A common voice (each doc reads like the others)
* A more friendly tone
* Easier navigation between information

Our `docs/` folder has two sub-folders:

* Conceptual
* Development

From there, I have created an aspirational table of contents for all the content I think we should document. It looks like this:

```
Overview
-Overview
-Getting Started
-Glossary

Conceptual
-Conceptual Documentation
-Substrate Core
--Parity Scale Codec
--Consensus
--Executor
--Networking
--Storage
-Substrate Nodes
--Client
--Light Client
--Extrinsics
--RPC
--Transaction Pool
-Substrate Runtime
--Runtime Overview
--Substrate Runtime Module Library
--Substrate Smart Contracts
--Runtime Primitives
-Keyring
--Cryptography
--Session Keys
--Subkey

Development
-Developer Documentation
-Runtime Modules
--Overview
--Runtime Execution
--Runtime Macros
--Runtime Storage
--Runtime Events
--Runtime Errors
--Runtime Origin
--Runtime Configuration Traits
--Transaction Weight
--Runtime Tests
-Build
--Prerequisites
--Purge Chain
--Runtime Upgrades
-Front End
--Polkadot-JS
--JSON RPC
-Deployment
--Chain Specification
--Account Prefix
--Boot Nodes
--Disabling Features of Your Runtime
```

Each one of these files in this PR has suggested information which should be covered by this doc.


4meta5:
* almost everything in the `Substrate Core` section already has reference docs which are better suited for the target audience (protocol devs *building* with Substrate) -- doesn't seem to make a lot of sense in the short-term to repeat these docs (anyone trying to understand the design of the system will read the w3f papers and look at the reference docs)
* should the `keyring` section be in conceptual or development? `subkey` should probably be in `Development`
* `Runtime Configuration Traits` should be `Runtime Types` because the `Trait` is used for specifying the required form when configuring types for the module in the runtime
* don't think we should document `Transaction Weight`, at least not at first. The relay chain is using the current transaction weight design to manage costs, but this may not be the appropriate model for every parachain (consider the motivation behind the `feeless token tutorial`)
* is `Developer Documentation` the docs on the docs? I feel like this could be kept separately, but I don't have a strong opinion. The longer this table of contents is, the harder it will be to find what they're looking for (and the easier it is to get distracted and give up)

Otherwise, this looks good. We should discuss prioritization soon. It would be nice to identify (agree on)
1. what sections need to be done first
2. who wants to do which sections
3. establish links between conceptual - documentation so the quality (linking norms?) are consistent across all docs instead of some docs linking frequently and others neglecting to do so

I'd prefer to focus on runtime configuration and module development, the `Development/Runtime Modules` section. I am also open to helping with `Conceptual/Substrate Node` section and can pick up other spare tasks as needed.

JoshOrndorff:
If we want to contribute to this effort (eg by writing docs) how should we do it? Make another PR against `shawntabrizi:v2.0`?

shawntabrizi:
Closed in favor of #252

PR #253, opened 2019-09-24
Conceptual: SCALE Codec

shawntabrizi:
- [x] Get examples for all formats


bkchr:
Why not just link to the SCALE codec reference at web3? And add links to the implementations.

shawntabrizi:
@bkchr I have added a reference to the Polkadot Runtime Environment specification. As we talked about offline, most other docs point to this page, so we will keep it up to date.

docs/conceptual/core/codec.md
```diff
@@ -1,13 +1,146 @@
...
 
-* Why did we invent our own codec?
+Substrate uses the [`parity-scale-codec`](https://github.com/paritytech/parity-scale-codec), a Rust implementation of the SCALE Codec. This library and the SCALE codec is advantageous for Substrate and blockchain systems because:
 
-* What implementations exist?
\ No newline at end of file
+* It does not use Rust STD, and thus can compile to Wasm.
+* It is built to have great support in Rust for deriving codec logic for new types: just add `#[derive(Encode, Decode)]`.
+
+It is important to define the encoding scheme used on Substrate rather than reuse an existing Rust codec library because this codec needs to be re-implemented on other platforms and languages that want to support interoperability.
```
JoshOrndorff: I don't really understand this claim. Isn't reuse on other platforms and in other languages a reason _to_ use an already-popular codec? I thought parity invented a new one to be super lightweight. Don't remember why u thought that though.

docs/conceptual/core/codec.md
```diff
@@ -1,13 +1,146 @@
...
+The SCALE (Simple Concatenated Aggregate Little-Endian) Codec is a lightweight, efficient, binary serialization and deserialization codec.
 
-* How does it work?
+It is designed for high-performance, copy-free encoding and decoding of data in resource-constrained execution contexts, like the [Substrate runtime](conceptual/runtime/index.md). It is not self-describing in any way and assumes the decoding context has all type knowledge about the encoded data.
 
-* Examples
+## SCALE for Substrate
 
-* Why did we invent our own codec?
+Substrate uses the [`parity-scale-codec`](https://github.com/paritytech/parity-scale-codec), a Rust implementation of the SCALE Codec. This library and the SCALE codec is advantageous for Substrate and blockchain systems because:
```
JoshOrndorff:
```diff
- is advantageous
+ are advantageous
```

docs/conceptual/core/codec.md
```diff
@@ -1,13 +1,146 @@
...
+
+### Fixed-width Integers
+
+Basic integers are encoded using a fixed-width little-endian (LE) format.
+
+#### Example
+
+- `signed 8-bit integer 69`: `0x45`
+- `unsigned 16-bit integer 42`: `0x2a00`
+- `unsigned 32-bit integer 16777215`: `0xffffff00`
```
JoshOrndorff: Are 64 but also supported? I don't think you need an example of it, but probably at least mention it. Actually I think up to 512 bits are supported.

docs/conceptual/core/codec.md
```diff
@@ -1,13 +1,146 @@
...
+### Boolean
+
+Boolean values are encoded using the least significant bit (LSB) of a single byte.
+
+#### Example
+
+- `boolean false`: `0x00`
+- `boolean true`: `0x01`
+
+### Compact/General Integers
```
JoshOrndorff: This should go right after (or right before) fixed width integers.

docs/conceptual/core/codec.md
```diff
@@ -1,13 +1,146 @@
...
+One or zero values of a particular type. Encoded as:
+
+- `0x00` if it is `None` ("empty" or "null").
+- `0x01` followed by the encoded value if it is `Some`.
+
+As an exception, in the case that the type is a boolean, then it is always one byte:
+
+- `0x00` if it is `None` ("empty" or "null").
+- `0x01` if it is the `false` value.
+- `0x02` if it is the `true` value.
```
JoshOrndorff: Should this also be mentioned in the Boolean section? Maybe putting these sections side by side would also help clarify that?

docs/conceptual/core/codec.md
```diff
@@ -1,13 +1,146 @@
...
+* JavaScript: [`polkadot-js/api`](https://github.com/polkadot-js/api)
+
+## Next Steps
+
+### Learn More
+
+TODO
+
+### Examples
+
```
JoshOrndorff: I'm comfortable omitting this since you have examples throughout.

docs/conceptual/core/codec.md
shawntabrizi: I think the thing I am trying to get across is that _we_ control the standard used here. Basti was telling me that there are probably already existing codecs which are as light weight as this, and implemented in rust.

However, those codecs may not have a well defined specification (so that they can be implemented in other languages), and may change or break in some way outside of our control.

We define the spec in the Polkadot Runtime Environment specification, and thus we control how the codec evolves over time if needed.

Does that make sense? Should I just say this instead?

docs/conceptual/core/codec.md
shawntabrizi: Yeah, we can remove when we do a "final pass" on next steps at the end of writing most the docs.

Otherwise, it might be cool to just point to some of the in-code tests we have for the scale codec to see how it is pracitcally used.

Another good example might be to simply show how you can call the `.encode()` function on items to turn it into it's scale encoded format.

docs/conceptual/core/codec.md
shawntabrizi: Good suggestion. I moved Options to be right after Boolean. I didnt not mention anything explicitly, I think this should be reasonable enough. (Since option really is a different type than anything it wraps)

I wonder if practically speaking, the spec could have had `None` be `0x02` so that `false` and `true` stay the same byte representation.

docs/conceptual/core/codec.md
shawntabrizi: moved

docs/conceptual/core/codec.md
shawntabrizi: I think there are no limits with regard to fix-width integers. They are represented just as anyone would expect. You add more bytes, you get a bigger fixed width number.

shawntabrizi: Would you still want to mention something in this case?

JoshOrndorff: Oh good point. The codec can support any side integer. I was thinking about what was implemented. So u guess the question is whether to mention that explicitly. Your call on that.

docs/conceptual/core/codec.md
JoshOrndorff: Encoding and test examples are going to use a specific representation. I'm fine with that though. Would be cool to show examples in s couple different languages.

shawntabrizi: It would be from Substrate, which makes sense in the context of Substrate docs eh?

docs/conceptual/core/codec.md
```diff
@@ -1,13 +1,178 @@
...
 
-* Why did we invent our own codec?
+Substrate uses the
+[`parity-scale-codec`](https://github.com/paritytech/parity-scale-codec), a Rust
+implementation of the SCALE Codec. This library and the SCALE codec are
+advantageous for Substrate and blockchain systems because:
 
-* What implementations exist?
\ No newline at end of file
+* It does not use Rust STD, and thus can compile to Wasm.
```
bkchr: Serde can be compiled as well without `std`^^

shawntabrizi: I think it is still fair to call out no? I don't assume all codec libraries can compile with `no_std` right?

This one can :)

kianenigma: I have heard that our main problem with serde, that you could name here, is a sh*t ton of boilerplate that it produces and bloats the size of the binary apparently?

docs/conceptual/core/codec.md
```diff
@@ -1,13 +1,181 @@
...
+
+`(3, false)`: `0x0c00`
+
+### Enumerations (tagged-unions)
+
+A fixed number of variants, each mutually exclusive and potentially implying a
+further value or series of values.
+
+Encoded as the first byte identifying the index of the variant that the value
+is. Any further bytes are used to encode any data that the variant implies.
```
thiolliere: would it worth saying: thus no more than 256 variants are supported

docs/conceptual/core/codec.md
4meta5:
```suggestion
* it is lightweight relative to generic serialization frameworks like [serde](https://serde.rs/), which add significant boilerplate that often bloats the size of the binary
```

docs/conceptual/core/codec.md
```diff
@@ -1,13 +1,181 @@
...
+- `unsigned integer 42`: `0xa8`
+- `unsigned integer 69`: `0x1501`
+
+Error:
+
+- ~~`0x0100`: Zero encoded in mode 1~~
+
+### Boolean
+
+Boolean values are encoded using the least significant bit (LSB) of a single
```
4meta5: No need to introduce acronyms unless we refer to the concept again with the acronym. If I'm missing some use of LSB elsewhere in this doc, ignore this...
```suggestion
Boolean values are encoded using the least significant bit of a single
```

docs/conceptual/core/codec.md
shawntabrizi: updated

shawntabrizi:
@4meta5 @thiolliere Nits addressed

PR #257, opened 2019-09-27
Update private network tutorial to master branch

JoshOrndorff:
This PR updates the private network tutorial to work with Substrate master (specifically the commit that Substrate Package v1.5.0 uses). I'm happy with this tutorial in the sense that it works as advertised and works as well as the old one did.

I did still hope to do the following though, but these items are slightly blocked
- [ ] Write a explicitly against Package v1.5.0. Blocked because package doesn't include subkey
- [ ] Link to a dedicated how-to-build-the-node page. Blocked because no such page exists in 2.0
- [ ] Split into multiple pages. Not necessarily totally blocked, but I want to have the text more finalized, and see an example of how to do it first.


docs/tutorials/start-a-private-network-with-substrate.md
```diff
@@ -9,29 +9,33 @@ You will need:
...
-git clone -b v1.0 https://github.com/paritytech/substrate
+# Download the Substrate code
+git clone https://github.com/paritytech/substrate
 cd substrate
+
+# Switch to the proper commit
+git checkout 7d7e74fb
+
+# Install rust prerequisites
+curl https://getsubstrate.io -sSf | bash -s -- --fast
```
kaichaosun: I'd like to put this step in the first place, since it may take a long time.

docs/tutorials/start-a-private-network-with-substrate.md
```diff
@@ -145,77 +118,107 @@ yarn
...
+      "5GoNkf6WdbxCFnPdAnYYQyCjAKPJgLNxXwPjwTh6DGg6gN3E",
+      1
+    ]
+  ]
+},
+
 ```
 
-All we need to do is change the authority addresses listed (currently Alice and Bob) to our own addresses that we generated in the previous step. A single person should do these steps and share the resulting file. with their fellow validators. (Because Rust -> Wasm builds aren't "reproducible", each person will get a slightly different Wasm blob which will break consensus if each participant generates the file themselves.)
+All we need to do is change the authority addresses listed (currently Alice and Bob) to our own addresses that we generated in the previous step. The sr25519 addresses go in the `babe` section, and the ed25519 addresses in the `grandpa` section. You may add as many validators as you like.
```
kaichaosun: This is a wiki in polkadot protocol has some detail: https://wiki.polkadot.network/en/latest/polkadot/learn/keys/
Not sure if we want to link it here.

docs/tutorials/start-a-private-network-with-substrate.md
```diff
@@ -9,29 +9,33 @@ You will need:
...
 1. Compile a Substrate node from source code.
-2. Generate ed25519 key-pairs for use as a network authority
+2. Generate ed25519 and sr25519 key-pairs for use as a network authority
 3. Create and edit a chainspec json file
 
 
 Compiling the Tools
 -----------------
-Substrate is an open source framework for building your own blockchain nodes. It can be used for a very wide variety of applications.
+Substrate does not (yet) offer binary installation packages, so it must be compiled from source, which can be a time-consuming process. Commits go into the [Substrate repository](https://github.com/paritytech/substrate) regularly and it is wise that everyone participating in this network have the same version of substrate to guarantee success. In practice similar but not identical versions will often work, but banking on that is a recipe for frustration. In this tutorial, we'll be using commit `7d7e74fb` for more stability.
```
kaichaosun: Shall we share some context about why choose this commit, so users are more confident about their practice.

docs/tutorials/start-a-private-network-with-substrate.md
```diff
@@ -9,29 +9,33 @@ You will need:
...
+
+# Switch to the proper commit
+git checkout 7d7e74fb
+
+# Install rust prerequisites
+curl https://getsubstrate.io -sSf | bash -s -- --fast
 ```
 
-The repository includes a utility called `subkey` which we'll optionally  use to generate and inspect keypairs. Let's compile it first. The `--force` option means that we'll install this version over any previously installed versions.
+The repository includes a utility called `subkey` which we'll optionally use to generate and inspect keypairs. Let's compile it first. The `--force` option means that we'll install this version over any previously installed versions.
```
kaichaosun:
```suggestion
The repository includes a utility called `subkey` which we'll optionally use to generate and inspect keypairs. Let's compile it first. The `--force` option means that we'll install this version by overriding the previously installed version.
```

docs/tutorials/start-a-private-network-with-substrate.md
HCastano: I'd also like to see more about what this is downloading / doing, without having to go to the script myself. This is important since we're piping into bash, which is generally a security concern.

docs/tutorials/start-a-private-network-with-substrate.md
```diff
@@ -9,29 +9,33 @@ You will need:
...
+Now let's compile the actual blockchain node that we'll be running. Because Substrate is a framework, most real-world blockchains that use it will write custom runtime code. There are [other tutorials](/tutorials/) that cover that process in detail. Luckily the Substrate repository itself already comes with two ready-to-run node environments. The first lives in the `node` directory and includes many features to be a practical blockchain. In fact it looks quite similar to [Polkadot](https://polkadot.network) which is also built on Substrate. The second is a more minimal runtime that lives in the `node-template` directory. We'll be using the node template in this tutorial because of its simplicity and because it is the usual starting point when writing your own custom runtime.
```
HCastano: What, aside from the "simplicity" makes `node-template` the usual starting point of writing a runtime?

docs/tutorials/start-a-private-network-with-substrate.md
```diff
@@ -90,43 +64,42 @@ cd ..
...
+2019-09-26 15:42:10 Starting BABE Authorship worker
+2019-09-26 15:42:15 Idle (0 peers), best: #0 (0x8855…e564), finalized #0 (0x8855…e564), ⬇ 0 ⬆ 0
+2019-09-26 15:42:20 Idle (0 peers), best: #0 (0x8855…e564), finalized #0 (0x8855…e564), ⬇ 0 ⬆ 0
 ```
 
 A few lines to take note of:
-`Node name: AlicesNode` shows the human-readable name you specified or an autogenerated one
-`Local node identity is: QmNTx7qYd5JiasUTzjaAbTCUeH5ot9xMjRePRjkSEdD7MY` shows the node ID that Bob will need when booting from Alice's node
-`Listening for new connections on 127.0.0.1:9944.` shows that the node is listening for RPC connections on port 9944. By default only connections from localhost will be accepted, but you can change that behavior by specifying `--ws-external`. You can change the port by specifying `--ws-port 12345`.
+`Local node identity is: QmViTYm8Yr4yneyRGov9L87vLeVLoySFduVRNQ1d33WdXN` shows the node ID that Bob will need when booting from Alice's node
```
HCastano: A nitpick for consistency.

```suggestion
`Local node identity is: QmViTYm8Yr4yneyRGov9L87vLeVLoySFduVRNQ1d33WdXN` shows the node ID that Bob will need when booting from Alice's node.
```

docs/tutorials/start-a-private-network-with-substrate.md
```diff
@@ -145,77 +118,107 @@ yarn
...
 ----------------------
-Now that we know the fundamentals and the command line options, it's time to generate our own keys rather than using the well-known Alice and Bob keys.
-
-Each person who wants to participate in the blockchain can generate their own key using the [subkey tool](ecosystem/subkey) or the Polkadot JS Apps UI that we used previously. In this tutorial, we'll use the UI.
+Now that we know the fundamentals and the command line options, it's time to generate our own keys rather than using the well-known Alice and Bob keys. Each person who wants to participate in the blockchain can generate their own key using the [subkey tool](ecosystem/subkey). Be sure to record all of the output from this section as you will need it later.
 
-Navigate to the "Accounts" tab, and choose "Add Account". For this kind of session key, we'll need to select Edwards Curve near the bottom, and copy-paste the public address and the raw seed.
-
-![Screenshot of key generation](https://cdn.discordapp.com/attachments/559788157239951376/583363719753367567/unknown.png)
+First generate a mnemonic and see the sr25519 key and address associated with it.
```
HCastano: How do you feel about "highlighting" they key types? Also, I think giving some context while generating the keys will help the reader.

```suggestion
First generate a mnemonic and see the `sr25519` key and address associated with it. This key will be used by BABE for block production.
```

docs/tutorials/start-a-private-network-with-substrate.md
```diff
@@ -145,77 +118,107 @@ yarn
...
+```
+$ subkey generate
+Secret phrase `keep matrix knee meat awake frown rubber position federal easily strategy inhale` is account:
+  Secret seed: 0xb5d5cda89e139aecb67181e11d6d2d90a0cc80106afa035ab19264af7b5e5c0b
+  Public key (hex): 0x8ed5f822065e5824d3e37d9ea36a81eacb98ff1a6fa04bb87d2fa4915e9ed147
+  Address (SS58): 5FHzDem7A5aAq79tuEN9xJuNPXiYfmRQamhumTuqu6i57BuU
+```
 
-When you have generated a key pair, store the entire output somewhere safe. Share your public address with the other validators who will be in your network. When everyone in the network knows everyone else's addresses, you can proceed to launch your own chain.
+Now see the ed25519 key and address associated with the same mnemonic
```
HCastano:
```suggestion
Now see the `ed25519` key and address associated with the same mnemonic. This key will be used by GRANDPA for block finalization.
```

docs/tutorials/start-a-private-network-with-substrate.md
```diff
@@ -145,77 +118,107 @@ yarn
...
+```
+$ subkey --ed25519 inspect "keep matrix knee meat awake frown rubber position federal easily strategy inhale"
+Secret phrase `keep matrix knee meat awake frown rubber position federal easily strategy inhale` is account:
+  Secret seed: 0xb5d5cda89e139aecb67181e11d6d2d90a0cc80106afa035ab19264af7b5e5c0b
+  Public key (hex): 0xfe68fdff17960cb8d45d861396a64d4086997353849403ee3352996ec68ff4af
+  Address (SS58): 5HpHD5YseSWbHfni43Zm2SjtpyqSVmuaNhKkVmivp8L93Trs
+```
 
-Start Your Private Blockchain
+Creating the chainspec
```
HCastano:
```suggestion
Creating the Chainspec
```

docs/tutorials/start-a-private-network-with-substrate.md
```diff
@@ -145,77 +118,107 @@ yarn
...
 
-### Creating the chainspec
 Rather than writing our chainspec completely from scratch, we'll just make a few modifications to the the one we used before. To start we need to export the chainspec to a json file. Remember, further details about all of these commands are available by running `node-template --help`.
 
 ```bash
 ./target/release/node-template build-spec --chain=local > customSpec.json
 ```
 
-The file we just created contains several fields, and one can learn a lot by exploring them. By far the largest field is a single hex number that is the Wasm binary of our runtime. It is what you built earlier when you ran the `build.sh` script. Learn more about why it's useful to have that info on-chain from [Gav's web3 summit demo](https://www.youtube.com/watch?v=0IoUZdDi5Is).
+The file we just created contains several fields, and one can learn a lot by exploring them. By far the largest field is a single hex number that is the Wasm binary of our runtime. It is part of what you built earlier when you ran the `cargo build` command. Having this inforamtion on-chain allows supports forkless upgrades.
```
HCastano:
```suggestion
The file we just created contains several fields, and one can learn a lot by exploring them. By far the largest field is a single hex number that is the Wasm binary of our runtime. It is part of what you built earlier when you ran the `cargo build` command. Having this information on-chain allows supports forkless upgrades.
```

docs/tutorials/start-a-private-network-with-substrate.md
```diff
@@ -145,77 +118,107 @@ yarn
...
 
-The portion of the file we're interested in is the authorities which looks like this
+The portion of the file we're interested in is the babe authorities (used for creating blocks) and the grandpa authorieits (used for finalizing blocks). That section looks like this
```
HCastano:
```suggestion
The portion of the file we're interested in is the BABE authorities (used for creating blocks) and the GRANDPA authorities (used for finalizing blocks). That section looks like this
```

docs/tutorials/start-a-private-network-with-substrate.md
HCastano: I think it would be good to link it. Something like, "If you're interested in reading more about the key pairs, go to ...".

docs/tutorials/start-a-private-network-with-substrate.md
```diff
@@ -225,27 +228,51 @@ Once the chainspec is prepared, convert it to a "raw" chainspec. The distinction
...
-  --key "enroll mechanic ... embody" \
   --port 30333 \
   --telemetry-url ws://telemetry.polkadot.io:1024 \
   --validator
+  --name MyNode
 ```
 Here are some differences from when we launched as Alice.
 * I've omitted the `--base-path` flag, so Substrate will use a default location. You're still free to specify your own base path.
-* For the key, I've used the mnemonic phrase that we generated previously. This must be a phrase that corresponds to one of the addresses in your chainspec.
+* I've omitted the `--alice` flag instead we will insert our own custom keys into the keysotre through the RPC shortly.
```
HCastano:
```suggestion
* I've omitted the `--alice` flag. Instead we will insert our own custom keys into the keystore through the RPC shortly.
```

docs/tutorials/start-a-private-network-with-substrate.md
```diff
@@ -225,27 +228,51 @@ Once the chainspec is prepared, convert it to a "raw" chainspec. The distinction
...
 * The `--chain` flag has changed to use our custom chainspec.
+* I've added the optional `--name` flag. You may use it to give your node a human-readable name in the telemetry UI.
+
+One your node is running, you will again notice that no blocks are being produced. At this point you can use the Apps UI to insert your keys into the keystore. NAvigate to the "Toolbox" tab and the "RPC Call" sub-tab. Choose "author" and "insertKey". The fields can be filled like this:
```
HCastano:
```suggestion
One your node is running, you will again notice that no blocks are being produced. At this point you can use the Apps UI to insert your keys into the keystore. Navigate to the "Toolbox" tab and the "RPC Call" sub-tab. Choose "author" and "insertKey". The fields can be filled like this:
```

docs/tutorials/start-a-private-network-with-substrate.md
JoshOrndorff: I totally agree about. The reason I didn't write that kind of stuff here is because I believe the plan is to have some unified install instructions that I should just link to from here.

JoshOrndorff: I agree. I didn't write that material here because I hope we can have setup-your-developer-environment instructions on a dedicated page. That's my second checkbox in the description.

docs/tutorials/start-a-private-network-with-substrate.md
JoshOrndorff: I plan to rewrite this part in terms of Substrate Package. I'm trying to figure out the best way to get users a compatible version of subkey. This is my first checkbox. See also https://github.com/substrate-developer-hub/substrate-package/issues/25

docs/tutorials/start-a-private-network-with-substrate.md
JoshOrndorff: I think the fact that it was designed to be a good starting point is why we usually start with it. Is there some text you'd like added here?

docs/tutorials/start-a-private-network-with-substrate.md
```diff
@@ -145,77 +118,107 @@ yarn
...
+    ]
+  ]
+},
+
 ```
 
-All we need to do is change the authority addresses listed (currently Alice and Bob) to our own addresses that we generated in the previous step. A single person should do these steps and share the resulting file. with their fellow validators. (Because Rust -> Wasm builds aren't "reproducible", each person will get a slightly different Wasm blob which will break consensus if each participant generates the file themselves.)
+All we need to do is change the authority addresses listed (currently Alice and Bob) to our own addresses that we generated in the previous step. The sr25519 addresses go in the `babe` section, and the ed25519 addresses in the `grandpa` section. You may add as many validators as you like.
+
+> A single person should do these steps and share the resulting file with their fellow validators.
```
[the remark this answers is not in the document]
JoshOrndorff: It's not a quote we're just abusing markdown to make it render in a yellow box.

docs/tutorials/start-a-private-network-with-substrate.md
shawntabrizi: Phrases like "but banking on that is a recipe for frustration" probably does not translate well. See our contribution guidelines:

> Avoid using uncommon words, highly technical words, or jargon that users might not understand. Also, avoid using idioms such as "that's the way the cookie crumbles", while it might make sense to you, it may not translate well into another language. Keep in mind that a lot of users are non-native English speakers.

IMO too much detail here around versioning. I would have the text assume it does work, and a small note saying if it doesn't to open an issue.

shawntabrizi:
I think a lot of the awkwardness of the setup can be avoided if instead of using Subkey we used the Polkadot JS API, Polkadot JS Apps, or even a new module we create for `substrate-front-end-template`.

I would want to review this again when this gets changed over.

JoshOrndorff:
I'm a big fan of Apps, but we can't use it for key generation because it doesn't provide a raw pubkey which is needed for the `insertKey` call.

Using a new component in frontend-template might be okay but it has tow distinct disadvantages
1. It requires introducing the frontend template in this tutorial where it is otherwise not used.
2. It hasn't been written yet.

For those reasons, I prefer to stick with subkey.

shawntabrizi:
> it doesn't provide a raw pubkey 

I am not sure this is accurate, but we should talk offline.

Okay, so do you want to merge this (since it is as good as the old tutorial), and open another issue to continue to improve it? When items are not blocked.

JoshOrndorff:
My ToDos when I posted this issue were:
* Write a explicitly against Package v1.5.0. -- This won't happen because we decided not to include subkey in package. Instead I clarified why I chose the commit I did as recommended by @kaichaosun 
* Link to a dedicated how-to-build-the-node page. -- This won't happen because the new style guide requires that tutorials be self-contained.
* Split into multiple pages. -- This could still happen, but I'm not personally very motivated to do it, and it can be another PR.
```

> > it doesn't provide a raw pubkey
>
> I am not sure this is accurate, but we should talk offline.

If apps can generate keys and provide the raw pubkey, please say so here so we can all learn.

> Okay, so do you want to merge this

Yes, will merge now.

**Comment by shawntabrizi** [comment]
@JoshOrndorff Using the polkadot js API you can do things like:

```
m = util_crypto.mnemonicGenerate()
"cave green hair day gift winter gaze text salon crime refuse rate"
```

```
a = util_crypto.mnemonicToSeed("cave green hair day gift winter gaze text salon crime refuse rate")
Uint8Array(32) [ 203, 83, 237, 145, 22, 158, 107, 169, 175, 253, … ]
```

```
util_crypto.encodeAddress(s)
"5GfJWWUNib1HqbYwXsaA8Az4E1EVaHioJsL22gn3TaW7znzY"
```

This is available to a user right out of the `substrate-front-end-template` with our embedded command line thing, or you could embed the appropriate JS directly into your tutorial page.

JoshOrndorff:
That's a lot different than being able to do it in [polkadot-js apps](https://github.com/polkadot-js/apps).

I don't think it is preferable to introduce the user to using the API like that in this tutorial because
* While anyone can copy paste the commands, understanding it requires understanding javascript
* This is, so far, a non-coding tutorial
* It requires one of:
  * Installing the api locally (whole extra step from different repo, needs npm, etc)
  * Generating keys in a web page (Not good security practice)

shawntabrizi:
Is this what you think we can't do?

image

I mentioned somewhere, can't remember where, but for the purposes of the tutorial, we can tell the user that generating a key on the webpage is not a best practice, but for tutorial purposes.

I mean using Subkey on a computer that has seen the internet is not a good practice, but where do you draw the line?

The goal of your tutorial should get the user to accomplish _your_ task as quick and easy as possible. The task here isn't safe/secure key generation.

4meta5:
The second key generated has a typo where it says its `ed25519` but then the code block claims it is `sr25519` (in launching your chain section)

JoshOrndorff:
> Is this what you think we can't do?
No. That's a raw seed. The seed is the entropy from which the key pair is generated. What I'm claiming Apps does not provide is the raw public key. We need it for [this API call](https://github.com/paritytech/substrate/blob/00ac65fca22652807e578326ad9f2fe6b169390a/core/rpc/api/src/author/mod.rs#L47)

shawntabrizi:
@JoshOrndorff 

Jaco was nice enough to make some custom UI which I think should solve the problem I mentioned here?

https://github.com/polkadot-js/apps/pull/1683

Take a look and see if you agree.

jacogr:
I'm open for suggestions as to making it easy, as per this comment https://github.com/polkadot-js/apps/pull/1683#issuecomment-537216976 the insertKey is cool, but it doesn't solve this problem. Additionally, we have the issue that we still require the user to concat all the stuff - this is just crazy and fraught with error.

So the apps UI adjusts based on the chain we connect to, so for instance for Polkadot, it shows claims, for generic assets it shows that transfer, etc. (Can do kitties for kitties, poe for poe, etc...)

Anyway, if there is a way to detect "this type of chain", we can have a simple UI that just allows you to create for the specific keys and actually send the set session keys as well. Suggestions welcome as to what and how.

shawntabrizi:
@jacogr in general, I think this tutorial is not targeted to any "specific" chain.

Its more about users who are still learning the basic basics of Substrate. In this case, they may have never written or touched a single line of code.

But they want to see what it would take to set up a small private network.

In this case, the Polkadot JS Apps does not provide them with the utilities to easily fill out the RPC call.

Solutions, for the purposes of this tutorial, would be things like:

* Show the Hex version of an address somewhere in the accounts section
* Have a utility in the tools which converts from SS58 to hex
* Allow for SS58 as an input into the RPC function
* etc...

Otherwise it really is just a lot for a user to handle, especially since they start asking "why?"

Like "why if I need the hex version of an account, do you show me some OTHER address?"

It is probably not the fault of the UI, but the fact that these RPCs and APIs seem to be designed with minimal customer thought in mind...

PR #261, opened 2019-09-30
Add (configurable) Hacktoberfest blast

gnunicorn:
Add a(n optional) blast on the top of the page to advertise cool things - in this case hacktoberfest:
image

image

image

----

It can even be configured: text, link, label and background can easily be set in the `siteConfig`, and all is turned off if the item `blast` isn't present (or set to `false`).

Please not: doccusaurus only allows customizations from v2.0, so the way this works is by us adding the element to the footer and then using css-magic to stick it before any other content and _move all the body_ down by that new element. It is not using any non-spec or heavy performance features but it simple isn't the way you'd do it properly...


gnunicorn:
Indeed forgot to add the file. Pushed.

> Could you also double check formatting on spacing?

How do you mean? I've not indented the inline-styles consciously, because it is one long js-string and they'd be included, but if you prefer it differently, I can do that.

I'd like to use the background for this time around. Solid color could still be used as a generic (and is built in for that purpose) on other occasions.

shawntabrizi:
> How do you mean? I've not indented the inline-styles consciously, because it is one long js-string and they'd be included, but if you prefer it differently, I can do that.

I would just do whatever auto formatting VS code recommends, but its not a big deal. Merge whenever you want.

PR #264, opened 2019-10-01
Fix Overflow from Footer

### LINKED ISSUE #260: width > 100%
For some reason some of our elements float out of the page width (even on very wide screens) on firefox. See the scroll-bar on the bottom. It is not visible what exactly is causing this, but it is present on all pages.

image

shawntabrizi:
Fixes #260 

CC @kaichaosun


website/core/Footer.js
```diff
@@ -132,37 +131,31 @@ class Footer extends React.Component {
...
-        <section className="row">
-          <div className="copyright">{this.props.config.copyright}</div>
-          <a className="item" href="https://www.parity.io/privacy/">Privacy Policy</a>
-          <a className="item" href="#" id="cookie-settings">Cookie Settings</a>
-          <script
-            dangerouslySetInnerHTML={{
-              __html: `
+        <section className="copyright">{this.props.config.copyright}</section>
+
+        <script
```
kaichaosun: Will it be better if put the script near the element?

shawntabrizi: it doesn't matter right?

I could put a comment on the script saying they are related

shawntabrizi: I moved the script right into the `a` tag and also put a note :)

PR #267, opened 2019-10-01
Add hacktoberfest back

kaichaosun:
Notes:
The header will not be fixed. Is this ok? @shawntabrizi

PR #269, opened 2019-10-01
Use margin to show blast

gnunicorn:
Replaces transition on body with margin, which does not break the fixed elements. It however, does influence sticky-positioning and thus this also contains a fix for the menus to align properly. After it, all looks like this (notice how we are both fixed, the menus and klaro float correctly - also when scrolling)

image

superseeds #267


website/core/Footer.js
```diff
@@ -49,6 +46,19 @@ class Blast extends React.Component {
   color: white;
   text-decoration: underline;
 }
+
+body {
+  // let's give it a minimal animation
+  transition: ease 1.5s;
```
gnunicorn: this adds a tiny animation to ease in the top at the load of the page. Not sure we want to have this, considering that one can't "x" it away and it moves the content down by 50px on every page load. Might want to switch it off, even if that means there is a short flicker for re-render on slower/delayed loads of the footer.

gnunicorn:
Sure, I don't mind. Didn't change any of that in this PR though ;) .

Will do tomorrow.

shawntabrizi:
Right, but to me it wasn't so noticeable when it scrolled away during navigation of the page.

Now that it is always there at the top, you feel it more (imo)

It is approved, so merge in whatever you think is right. Just wanted to voice an opinion

PR #271, opened 2019-10-02
Increase Blast `z-index`

shawntabrizi:
Solves this:
image


shawntabrizi:
@gnunicorn c'mon dude. It has to be over 9000...

PR #275, opened 2019-10-02
Replace double quote with single quotes in JS for PoE tutorial

HCastano:
cc @shawntabrizi

### DIFF
--- a/docs/tutorials/creating-your-first-substrate-chain/front-end.md
+++ b/docs/tutorials/creating-your-first-substrate-chain/front-end.md
@@ -20,11 +20,11 @@ substrate-front-end-template
 |
 +-- src
 |   |
-|   +-- index.jsx
+|   +-- index.js
 |   |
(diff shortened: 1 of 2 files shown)


JoshOrndorff:
This is so the linting passes when it is pasted into the front-end-template, right?

shawntabrizi:
yes, otherwise `yarn start` I believe pings the user.

I was hoping that linting would be more passive. Like before I would just auto-format the files I would touch when working on the `substrate-front-end-template`.

However, causing compilation warnings and stuff is a bit more than I would have personally wanted.

JoshOrndorff:
I think it's good that code we're promoting does pass linting, so I'm glad this PR went in. But I agree with you that users who are just playing around and learning don't need to be burdened by an opinionated linter.

Maybe move this convo to this [front-end-template issue](https://github.com/substrate-developer-hub/substrate-front-end-template/issues/54)?

PR #277, opened 2019-10-02
Ben remove blast

gnunicorn:
based off #276

this removes the blast and all the mess we did with the css and instead opt for a slightly highlighted menu item:

image

Merge whenever we find yet another issue with the blast - but only if so.


shawntabrizi:
Replaced with #300

PR #278, opened 2019-10-06
Add `getting-started` docs to `next`

shawntabrizi:
Only thing missing here is "Next Steps" items, which we can add when we do final pass of the "next" documentation.

Uses _no_ scripts... although we might want to consider including the `getsubstrate.io` script since it is really nice for Mac/Linux users.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -6,7 +6,7 @@ This page has everything you need to get started building with Substrate.
 
 ## Get Help
 
-Substrate is a bleeding edge blockchain development framework, and as a result you may encounter some breaking changes or issues along the way.
+Substrate is a bleeding edge blockchain development framework, and as a result, you may encounter some breaking changes or issues along the way.
 
 Know that we are very happy to help you get started building with Substrate, so if you encounter any issues, simply reach out to us by:
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -207,7 +207,7 @@ Connect to the front-end at [`localhost:8000`](http://localhost:8000).
 
 ## Interact with Substrate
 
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -2,4 +2,279 @@
...
+https://github.com/substrate-developer-hub/substrate-front-end-template
+```
+
+Install the node dependencies with:
+
+```bash
+yarn install
+```
+
+Run the front end with:
```
4meta5: For consistency, should we use front-end or front end? See line 184, "front-end"
```suggestion
Run the front-end with:
```

docs/getting-started.md
```diff
@@ -2,4 +2,279 @@
...
+
+```bash
+yarn start
+```
+
+Connect to the front-end at [`localhost:8000`](http://localhost:8000).
+
+> **Note:** You need to have a local Substrate node running to interact with this UI.
```
4meta5: Would be nice to link here to direction for getting a local Substrate node running.

docs/getting-started.md
```diff
@@ -2,4 +2,279 @@
...
+
+## Interact with Substrate
+
+The fastest way to interact any local or public Substrate network it to visit Polkadot-JS Apps:
```
4meta5: > 
```suggestion
The fastest way to interact any local or public Substrate network is to visit Polkadot-JS Apps:
```

docs/getting-started.md
shawntabrizi: "front-end" is an adjective:
> front-end template

Describes the kind of template.

"front end" is a noun, as it is being used here.

shawntabrizi:
@4meta5 Nits addressed :)

PR #281, opened 2019-10-08
Conceptual Doc: Executor

shawntabrizi:
Here is a doc on the Substrate Executor which goes into details about Wasm and Native Runtime, and ultimately finishes with Runtime Upgrades.

### DIFF
--- a/docs/conceptual/core/executor.md
+++ b/docs/conceptual/core/executor.md
@@ -2,8 +2,91 @@
 title: Executor
 ---
 
-* Runtime Versioning
+The executor is responsible for dispatching and executing calls into the Substrate runtime.
 
-* Native VS Wasm Execution
+## Runtime Execution
 
-* Forkless Runtime Upgrades
+The Substrate runtime is compiled into a native executable and a WebAssembly (Wasm) binary.
+
+The native runtime is included as part of the node executable, while the Wasm binary is stored on the blockchain under a well known storage key.
+
+These two representations of the runtime may not be the same. For example: when the runtime is upgraded. The executor determines which version of the runtime to use when dispatching and executing calls.
+
+### Execution Strategy
+
+Substrate allows you to configure an execution strategy during different parts of the blockchain execution process. The strategies are:
+
+- `Native`: Execute with native build (if available, WebAssembly otherwise).
+- `Wasm`: Only execute with the WebAssembly build.
+- `Both`: Execute with both native (where available) and WebAssembly builds.
+- `NativeElseWasm`: Execute with the native build if possible; if it fails, then execute with WebAssembly.
+
+The default execution strategies for the different parts of the blockchain execution process are:
+
+- Syncing: `NativeElseWasm`
+- Block Import: `NativeElseWasm`
+- Block Construction: `Wasm`
+- Off-Chain Worker: `Native`
+- Other: `Native`
+
+### Wasm Execution
+
+The Wasm representation of the Substrate runtime is considered the canonical runtime. Because this Wasm runtime is placed in the blockchain storage, the network must come to consensus about this binary. Thus it can be verified to be consistent across all syncing nodes.
+
+The Wasm execution environment is more restrictive than the native execution environment, because it makes few assumptions about the target machine it will run on. For example, the Wasm runtime always executes in a 32-bit environment.
+
+For these reasons, the blockchain prefers to do block construction with the Wasm runtime even though Wasm execution is measurably slower than native execution. Wasm execution can help to ensure that block producers create valid blocks.
+
+### Native Execution
+
+For all other execution processes other than block construction, the native runtime is preferred since it is more performant. However, in any situation where the executor determines the native executable should not be run, the canonical Wasm runtime is chosen instead.
+
+## Runtime Upgrades
+
+The Substrate runtime is designed to be upgraded so that you can continue to iterate on your state transition function even after your blockchain goes live.
+
+### Runtime Versioning
+
+In order for the executor to be able to select the appropriate runtime execution environment, it needs to know the name and version of both the native and Wasm runtime.
+
+The runtime provides the following versioning properties:
+
+- `spec_name`: The identifier for the different Substrate runtimes.
+
+- `impl_name`: The name of the implementation of the spec. This is of little consequence for the node and serves only to differentiate code of different implementation teams.
+
+- `authoring_version`: The version of the authorship interface. An authoring node will not attempt to author blocks unless this is equal to its native runtime.
+
+- `spec_version`: The version of the runtime specification. A full-node will not attempt to use its native runtime in substitute for the on-chain Wasm runtime unless all of `spec_name`, `spec_version`, and `authoring_version` are the same between Wasm and native.
+
+- `impl_version`: The version of the implementation of the specification. Nodes are free to ignore this; it serves only as an indication that the code is different; as long as the other two versions are the same then while the actual code may be different, it is nonetheless required to do the same thing. Non-consensus-breaking optimizations are about the only changes that could be made which would result in only the `impl_version` changing.
+
+As mentioned above, the executor always verifies that the native runtime has the same consensus-driven logic before it chooses to execute it, independent of whether the version is higher or lower.
+
+> **Note:** The runtime versioning is manually set. Thus the executor can still make inappropriate decisions if the runtime version is misrepresented.
+
+### Forkless Runtime Upgrades
+
+Traditional blockchains require a [hard fork](https://en.wikipedia.org/wiki/Fork_(blockchain)) when upgrading the state transition function of their chain. This requires node operators to stop their nodes and manually upgrade to the latest executable. For distributed production networks, coordination of a hard fork upgrades can be a complex process.
+
+The culmination of the properties listed on this page allows for Substrate-based blockchains to perform "forkless runtime upgrades". This means that the upgrade of the runtime logic can happen in real time without causing a fork in the network.
+
+To perform a forkless runtime upgrade, Substrate uses existing runtime logic to update the Wasm runtime stored on the blockchain to a new version with new logic. This upgrade gets pushed out to all syncing nodes on the network as a part of the consensus process. Once the Wasm runtime is upgraded, the executor will see that the native runtime version no longer matches this new Wasm runtime. As a result, it will fall back to execute the canonical Wasm runtime instead of using the native runtime in any of the execution processes.
+
+## Next Steps
+
+### Learn More
+
+* Learn more about
+
+### Examples
+
+TODO
+
+### References
+
+* Check out the different [Execution Strategies](https://substrate.dev/rustdocs/master/substrate_service/config/struct.ExecutionStrategies.html).
+
+* Take a look at the different [Execution Strategy Options](https://substrate.dev/rustdocs/master/substrate_client/enum.ExecutionStrategy.html)
+
+* Review the [Runtime Version definition](https://substrate.dev/rustdocs/master/substrate_executor/struct.RuntimeVersion.html).
--- a/docs/conceptual/core/executor.md
+++ b/docs/conceptual/core/executor.md
@@ -15,7 +15,7 @@ binary is stored on the blockchain under a well known storage key.
 
 These two representations of the runtime may not be the same. For example: when
 the runtime is upgraded. The executor determines which version of the runtime to
-use when dispatching and executing calls.
+use when dispatching calls.
 
 ### Execution Strategy
 
@@ -49,7 +49,7 @@ consistent across all syncing nodes.
 
 The Wasm execution environment can be more restrictive than the native execution
 environment. For example, the Wasm runtime always executes in a 32-bit
-environment with a 4 GB memory limit.
+environment with a configurable memory limit (up to 4 GB).
 
 For these reasons, the blockchain prefers to do block construction with the Wasm
 runtime even though Wasm execution is measurably slower than native execution.
--- a/docs/conceptual/core/executor.md
+++ b/docs/conceptual/core/executor.md
@@ -24,7 +24,8 @@ execution environment should be used. This is controlled by the execution
 strategy, which can be configured for the different parts of the blockchain
 execution process. The strategies are:
 
-- `Native`: Execute with native build (if available, WebAssembly otherwise).
+- `Native`: Execute with native build (if available, WebAssembly otherwise); if
+  it fails, do not fallback to Wasm.
 - `Wasm`: Only execute with the WebAssembly build.
 - `Both`: Execute with both native (where available) and WebAssembly builds.
 - `NativeElseWasm`: Execute with the native build if possible; if it fails, then
@@ -52,7 +53,9 @@ environment with a 4 GB memory limit.
 
 For these reasons, the blockchain prefers to do block construction with the Wasm
 runtime even though Wasm execution is measurably slower than native execution.
-Wasm execution can help to ensure that block producers create valid blocks.
+Some logic executed in Wasm will always work in the native execution
+environment, but the same cannot be said the other way around. Wasm execution
+can help to ensure that block producers create valid blocks.
 
 ### Native Execution
 
@@ -73,7 +76,8 @@ In order for the executor to be able to select the appropriate runtime execution
 environment, it needs to know the `spec_name`, `spec_version` and
 `authoring_version` of both the native and Wasm runtime.
 
-The runtime provides the following [versioning properties](https://substrate.dev/rustdocs/master/sr_version/struct.RuntimeVersion.html):
+The runtime provides the following [versioning
+properties](https://substrate.dev/rustdocs/master/sr_version/struct.RuntimeVersion.html):
 
 - `spec_name`: The identifier for the different Substrate runtimes.
 
--- a/docs/conceptual/core/executor.md
+++ b/docs/conceptual/core/executor.md
@@ -2,26 +2,36 @@
 title: Executor
 ---
 
-The executor is responsible for dispatching and executing calls into the Substrate runtime.
+The executor is responsible for dispatching and executing calls into the
+Substrate runtime.
 
 ## Runtime Execution
 
-The Substrate runtime is compiled into a native executable and a WebAssembly (Wasm) binary.
+The Substrate runtime is compiled into a native executable and a WebAssembly
+(Wasm) binary.
 
-The native runtime is included as part of the node executable, while the Wasm binary is stored on the blockchain under a well known storage key.
+The native runtime is included as part of the node executable, while the Wasm
+binary is stored on the blockchain under a well known storage key.
 
-These two representations of the runtime may not be the same. For example: when the runtime is upgraded. The executor determines which version of the runtime to use when dispatching and executing calls.
+These two representations of the runtime may not be the same. For example: when
+the runtime is upgraded. The executor determines which version of the runtime to
+use when dispatching and executing calls.
 
 ### Execution Strategy
 
-Before runtime execution begins, the Substrate client passes to the executor an execution strategy. Substrate allows you to configure an execution strategy during different parts of the blockchain execution process. The strategies are:
+Before runtime execution begins, the Substrate client chooses which runtime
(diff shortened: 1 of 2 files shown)


docs/conceptual/core/executor.md
```diff
@@ -2,8 +2,91 @@
...
 
-* Forkless Runtime Upgrades
\ No newline at end of file
+The Substrate runtime is compiled into a native executable and a WebAssembly (Wasm) binary.
+
+The native runtime is included as part of the node executable, while the Wasm binary is stored on the blockchain under a well known storage key.
+
+These two representations of the runtime may not be the same. For example: when the runtime is upgraded. The executor determines which version of the runtime to use when dispatching and executing calls.
+
+### Execution Strategy
```
bkchr: Conceptually, this happens above the executor. The executor just gets the information on how a function should be executed.

docs/conceptual/core/executor.md
```diff
@@ -2,8 +2,91 @@
...
+- Block Import: `NativeElseWasm`
+- Block Construction: `Wasm`
+- Off-Chain Worker: `Native`
+- Other: `Native`
+
+### Wasm Execution
+
+The Wasm representation of the Substrate runtime is considered the canonical runtime. Because this Wasm runtime is placed in the blockchain storage, the network must come to consensus about this binary. Thus it can be verified to be consistent across all syncing nodes.
+
+The Wasm execution environment is more restrictive than the native execution environment, because it makes few assumptions about the target machine it will run on. For example, the Wasm runtime always executes in a 32-bit environment.
```
bkchr: It does not makes assumptions about the target machine. Wasm is a vm that dictates the environment, 32bit, max 4gb etc. But you can run the VM everywhere you want.

docs/conceptual/core/executor.md
```diff
@@ -2,8 +2,91 @@
...
+
+For these reasons, the blockchain prefers to do block construction with the Wasm runtime even though Wasm execution is measurably slower than native execution. Wasm execution can help to ensure that block producers create valid blocks.
+
+### Native Execution
+
+For all other execution processes other than block construction, the native runtime is preferred since it is more performant. However, in any situation where the executor determines the native executable should not be run, the canonical Wasm runtime is chosen instead.
```
bkchr: Hmm, this "determines", I would rewrite this. The executor does not uses the native when it is told or when the version of the native is not compatible with the requested runtime version.

docs/conceptual/core/executor.md
```diff
@@ -2,8 +2,91 @@
...
+
+## Runtime Upgrades
+
+The Substrate runtime is designed to be upgraded so that you can continue to iterate on your state transition function even after your blockchain goes live.
+
+### Runtime Versioning
+
+In order for the executor to be able to select the appropriate runtime execution environment, it needs to know the name and version of both the native and Wasm runtime.
```
bkchr:
```suggestion
In order for the executor to be able to select the appropriate runtime execution environment, it needs to know the `spec_name`, `spec_version` and `authoring_version` of both the native and Wasm runtime.
```

docs/conceptual/core/executor.md
```diff
@@ -2,8 +2,91 @@
...
+
+### Forkless Runtime Upgrades
+
+Traditional blockchains require a [hard fork](https://en.wikipedia.org/wiki/Fork_(blockchain)) when upgrading the state transition function of their chain. This requires node operators to stop their nodes and manually upgrade to the latest executable. For distributed production networks, coordination of a hard fork upgrades can be a complex process.
+
+The culmination of the properties listed on this page allows for Substrate-based blockchains to perform "forkless runtime upgrades". This means that the upgrade of the runtime logic can happen in real time without causing a fork in the network.
+
+To perform a forkless runtime upgrade, Substrate uses existing runtime logic to update the Wasm runtime stored on the blockchain to a new version with new logic. This upgrade gets pushed out to all syncing nodes on the network as a part of the consensus process. Once the Wasm runtime is upgraded, the executor will see that the native runtime version no longer matches this new Wasm runtime. As a result, it will fall back to execute the canonical Wasm runtime instead of using the native runtime in any of the execution processes.
```
bkchr: If you just update the `impl_version`, you can still use the native version, even after an upgrade.

docs/conceptual/core/executor.md
shawntabrizi: Hmm, is this true?

We are talking about the Execution strategy here or the version lookup?

In the code, I see that `executor/src/native_executor.rs` calls the function `can_call_with`

Which ultimately does the version lookup logic:

```
	pub fn can_call_with(&self, other: &RuntimeVersion) -> bool {
		self.spec_version == other.spec_version &&
		self.spec_name == other.spec_name &&
		self.authoring_version == other.authoring_version
	}
```

This is what I meant by it "deciding"

shawntabrizi: I think this topic is relevant, so I will simply specify that this happens from the client when it calls the executor. Is that okay?

docs/conceptual/core/executor.md
shawntabrizi: Dumb question, but can I run the Wasm on a machine which does not have at least these specifications? For example a device with only 1 GB of memory.

docs/conceptual/core/executor.md
shawntabrizi: I clarified to say:

```
Once the Wasm runtime is upgraded, the executor will see that the native runtime `spec_name`, `spec_version`, or `authoring_version` no longer matches this new Wasm runtime.
```

The alternative was to say "consensus-breaking logic", but this is probably less explicit.

docs/conceptual/core/executor.md
shawntabrizi: Updated

docs/conceptual/core/executor.md
shawntabrizi: I added "Before runtime execution begins, the Substrate client passes to the executor an execution strategy."

docs/conceptual/core/executor.md
bkchr: Yes, why not? If you would need 4gb, you would need to give 3gb of swap.
But substrate also does not gives the runtime 4gb of memory, this can be controlled. I think the default should be something around 65mb. But don't quote me on this number.

docs/conceptual/core/executor.md
bkchr: The executor does not know anything about execution strategies. Just if it should try to use native, not more.

shawntabrizi: > Before runtime execution begins, the Substrate client chooses which runtime execution environment should be used. This is controlled by the execution strategy, which can be configured for the different parts of the blockchain execution process. The strategies are: ...

Better? Again, I would want to confirm that you agree talking about the execution strategy stuff totally makes sense here, even though it happens above the executor.

I just need to make sure the wording is accurate and understandable.

docs/conceptual/core/executor.md
shawntabrizi: So then this is false:

> The Wasm execution environment can be more restrictive than the native execution environment. For example, the Wasm runtime always executes in a 32-bit environment with a 4 GB memory limit.

I could simply omit any "example" if it is not accurate.

docs/conceptual/core/executor.md
bkchr: Yeah that is okay. It is almost good. Could we just replace "choose" with "propose" or something similar. Because, even when you choose native, but the executor finds a version mismatch, it executes it in wasm.

docs/conceptual/core/executor.md
```diff
@@ -2,8 +2,147 @@
...
+The Wasm representation of the Substrate runtime is considered the canonical
+runtime. Because this Wasm runtime is placed in the blockchain storage, the
+network must come to consensus about this binary. Thus it can be verified to be
+consistent across all syncing nodes.
+
+The Wasm execution environment can be more restrictive than the native execution
+environment. For example, the Wasm runtime always executes in a 32-bit
+environment with a 4 GB memory limit.
+
+For these reasons, the blockchain prefers to do block construction with the Wasm
```
kianenigma: This sort of implies: 

> for the reason of wasm being much more limited 

 Then 

> the blockchain prefers to do block construction with the Wasm

While I suppose you are referring to the fact that wasm runtime is canonical and hence it must be used as the production reference, despite being slower.

docs/conceptual/core/executor.md
```diff
@@ -2,8 +2,147 @@
...
+The Substrate runtime is designed to be upgraded so that you can continue to
+iterate on your state transition function even after your blockchain goes live.
+
+### Runtime Versioning
+
+In order for the executor to be able to select the appropriate runtime execution
+environment, it needs to know the `spec_name`, `spec_version` and
+`authoring_version` of both the native and Wasm runtime.
+
+The runtime provides the following versioning properties:
```
kianenigma: I would add a link from here to https://github.com/paritytech/substrate/blob/f3a830d3e37d78a869c78c037cc2ac6f0c4e5ddd/node/runtime/src/lib.rs#L79 maybe (it for sure helped me understand better when I peaked at this code)

kianenigma: I assume better to do it from crates.parity.io

shawntabrizi: Added a link to here: https://substrate.dev/rustdocs/master/sr_version/struct.RuntimeVersion.html

docs/conceptual/core/executor.md
shawntabrizi: I don't think I mentioned anything about "slow" here.

I am simply saying that Wasm is more restrictive, as in something executed in Wasm will always work in Native, but not vice-versa.

Hence, to create fully compatible blocks with all nodes, you must use Wasm.

shawntabrizi:
Waiting for an approval here to merge

docs/conceptual/core/executor.md
```diff
@@ -2,8 +2,147 @@
...
+use when dispatching and executing calls.
+
+### Execution Strategy
+
+Before runtime execution begins, the Substrate client proposes which runtime
+execution environment should be used. This is controlled by the execution
+strategy, which can be configured for the different parts of the blockchain
+execution process. The strategies are:
+
+- `Native`: Execute with native build (if available, WebAssembly otherwise).
```
brunoffranca: Isn't this strategy equal to NativeElseWasm? I don't understand the difference.

docs/conceptual/core/executor.md
```diff
@@ -2,8 +2,147 @@
...
+### Wasm Execution
+
+The Wasm representation of the Substrate runtime is considered the canonical
+runtime. Because this Wasm runtime is placed in the blockchain storage, the
+network must come to consensus about this binary. Thus it can be verified to be
+consistent across all syncing nodes.
+
+The Wasm execution environment can be more restrictive than the native execution
+environment. For example, the Wasm runtime always executes in a 32-bit
+environment with a 4 GB memory limit.
```
brunoffranca: The 4 GB memory limit is useful information to have on other docs.

docs/conceptual/core/executor.md
```diff
@@ -2,8 +2,147 @@
...
+- `NativeElseWasm`: Execute with the native build if possible; if it fails, then
+  execute with WebAssembly.
+
+The default execution strategies for the different parts of the blockchain
+execution process are:
+
+- Syncing: `NativeElseWasm`
+- Block Import: `NativeElseWasm`
+- Block Construction: `Wasm`
+- Off-Chain Worker: `Native`
```
brunoffranca: This seems to contradict what is said on the high-level docs (https://substrate.dev/docs/en/overview/off-chain-workers), where it says that "With a sandboxed Wasm execution environment, off-chain workers are fully secure...". It seems to imply that off-chain workers are always executed with Wasm.

shawntabrizi: This is a good point, I will follow up.

docs/conceptual/core/executor.md
shawntabrizi: What other docs?

docs/conceptual/core/executor.md
shawntabrizi: The difference is that if the native execution fails, it does _not_ try Wasm execution.

You think it would help to write this explicitly?

brunoffranca: Ok, I get it now. Native only checks availability of the native build, while NativeElseWasm checks availability and correct execution.
Maybe writing it a bit more explicitly would help, they definitely seem the same on a first read.

docs/conceptual/core/executor.md
brunoffranca: Maybe the high level docs or the recipes. I'm just saying this because on one of my modules I have a Vec as a storage value. That Vec could reach 4 GB, knowing that there is a 4 GB limit I can try another solution. Or maybe adding checks to make sure that the Vec doesn't reach 4 GB.

brunoffranca: Now that I think about, having the information that it is a 32-bit environment is also important. A Vec can easily reach 2**32 elements. This would make the runtime panic. So, I would also need to add a check to prevent the Vec from reaching 2**32 elements.

bkchr: How can a vec easily reach 2^32 elements? You will run into problems much earlier. Such a huge state, that need to be synchronized between all nodes, will kill your blockchain already much earlier.

brunoffranca: Hmm, I will correct my previous position. If a vec reaches 2^32 elements, given that the minimum element size is 1 byte, it would panic because it would exceed the 4 GB limit.
But other than that, reaching 2^32 elements or 4 GB is not that hard in a blockchain. 2^32 is ~4 billion. Ethereum currently has 80 million addresses. That's only a difference of 50x, it's not hard to imagine that it will reach 4 billion over the next few years.
Also, I couldn't find the current size of Ethereum's state (and so correct me if I'm wrong), but with 80 million addresses it has to be bigger than 4 GB by now (an average size of 50 bytes per address would be enough to exceed 4 GB).

bkchr: No one says that the state can not be bigger, you can just not load 80 million addresses at once. That would not work from the memory perspective and also you would not have the computational power to load all these addresses in the small time frame you have for building a block.

kianenigma: yeah -- even with an infinitely large state, if there is a code that wants to load 4GB into a Vec, and that code fails, I don't really blame the system but rather the code. In the context of a blockchain that really should not happen. 

Maybe offchain which also runs in wasm could be a catch here.

brunoffranca: Hmm, ok. I think I misunderstood the memory limit. Then it's a limit on the working memory, not on the state size. If instead of using a Vec, I implement the same data structure using a StorageMap and a StorageValue, I don't have any limit on size or number of elements. Am I thinking correctly?

bkchr: Yes.

docs/conceptual/core/executor.md
shawntabrizi: The answer I got is:

> The offchain stuff needs to be compatible with both (like the runtime itself). I was told that main use cases for offchain workers will require a lot of performance though, so they might only be run natively. However I was sure that the strategy is NativeElseWasm, I think we can safely change it to be more inclusive, but if we actually introduce some heavy computations it might require the validators to actually run in native only.

So the answer is that we will update the config. Remember that even `Native` setting will run Wasm if Native is not available.

docs/conceptual/core/executor.md
shawntabrizi: @brunoffranca I think the thing you are asking for, and makes sense, is a separate page just on Wasm and the Wasm environment in Substrate.

Maybe the set of assumptions about Wasm in Substrate, limitations/settings, difference between Contract and Runtime and Offchain Worker, etc... Use of Wasmi and soon Lightbeam.

Maybe there is even a seperate Wasm section which has all these subsections.

Would that solve your ask?

docs/conceptual/core/executor.md
shawntabrizi: Updated

docs/conceptual/core/executor.md
brunoffranca: Well, my initial concern about the Wasm execution environment was unfounded. And even if it wasn't, the limits would only matter in edge cases (ie: my whole thing with the 4 GB Vec). So an entire page on the Wasm environment will probably not be useful for most people (who just want to learn how to create their own modules).
That being said, it would be something interesting to read. I always like to learn what I can about the Substrate core, even if I don't really interact with it.

docs/conceptual/core/executor.md
```diff
@@ -2,8 +2,151 @@
...
+The native runtime will only be used by the executor when it is chosen as the
+execution strategy and it is compatible with the requested runtime version. For
+all other execution processes other than block construction, the native runtime
+is preferred since it is more performant. In any situation where the native
+executable should not be run, the canonical Wasm runtime is executed instead.
+
+## Runtime Upgrades
+
+The Substrate runtime is designed to be upgraded so that you can continue to
+iterate on your state transition function even after your blockchain goes live.
```
4meta5: "iterate on your state transition function"? I was thinking of either swapping `iterate on` for `adjust` or making this change (and the above one)
```suggestion
upgrade your state transition function even after your blockchain goes live.
```

docs/conceptual/core/executor.md
```diff
@@ -2,8 +2,151 @@
...
+
+The native runtime will only be used by the executor when it is chosen as the
+execution strategy and it is compatible with the requested runtime version. For
+all other execution processes other than block construction, the native runtime
+is preferred since it is more performant. In any situation where the native
+executable should not be run, the canonical Wasm runtime is executed instead.
+
+## Runtime Upgrades
+
+The Substrate runtime is designed to be upgraded so that you can continue to
```
4meta5: see comment below
```suggestion
The Substrate runtime is designed to be flexible so that you can continue to
```

docs/conceptual/core/executor.md
shawntabrizi: I don't like using upgrade twice in a sentence. Feels weird right?

I would be fine with `adjust` but would want to understand the change.

shawntabrizi: with the change below it seems fine

4meta5: I *feel* like iterate is the wrong word here. [from mwebster](https://www.merriam-webster.com/dictionary/iteration): a procedure in which repetition of a sequence of operations yields results successively closer to a desired result. Maybe it does work...

That's why I'm more in favor of changing the position of "upgrade", but idrc...

docs/conceptual/core/executor.md
```diff
@@ -2,8 +2,151 @@
...
\ No newline at end of file
+The Substrate runtime is compiled into a native executable and a WebAssembly
+(Wasm) binary.
+
+The native runtime is included as part of the node executable, while the Wasm
+binary is stored on the blockchain under a well known storage key.
+
+These two representations of the runtime may not be the same. For example: when
+the runtime is upgraded. The executor determines which version of the runtime to
+use when dispatching and executing calls.
```
bkchr: For me dispatching and executing is the same.

docs/conceptual/core/executor.md
```diff
@@ -2,8 +2,151 @@
...
+
+### Wasm Execution
+
+The Wasm representation of the Substrate runtime is considered the canonical
+runtime. Because this Wasm runtime is placed in the blockchain storage, the
+network must come to consensus about this binary. Thus it can be verified to be
+consistent across all syncing nodes.
+
+The Wasm execution environment can be more restrictive than the native execution
+environment. For example, the Wasm runtime always executes in a 32-bit
```
bkchr: While the 4gb limit is correct, substrate uses a much smaller memory limit for the runtime. (it is configurable)

shawntabrizi: Where is this configuration?

bkchr: https://github.com/paritytech/substrate/blob/master/core/primitives/storage/src/lib.rs#L59 This can be set in the runtime storage. Should be default to `1024`.

PR #290, opened 2019-10-15
update `Creating a Runtime Module` tutorial

jimmychu0807:
@JoshOrndorff @shawntabrizi  Feel free to update/modify on top of this branch

Updates mostly consisted of:

- the 1st section `Create your Node` clone from substrate-node-template, instead of `substrate-new-node`
- module template clone from https://github.com/substrate-developer-hub/substrate-module-template, instead of Shawn personal github repo.
- some path update, and the adding description on path relationship between node-template and module-template, and which `Cargo.toml` we are talking about.
- Don't recommend the v1.0 branch
- Added a next step section at the end

### DIFF
--- a/docs/tutorials/creating-a-runtime-module.md
+++ b/docs/tutorials/creating-a-runtime-module.md
@@ -336,5 +336,24 @@ code in their runtime's `Cargo.toml` files and updating their runtime's `lib.rs`
 
 In this tutorial, you've learned to create a runtime module as a Rust crate.
 You're now ready to replace the template logic with your own. If you're not sure
-what to code, check out [one of our tutorials](https://substrate.dev/en/tutorials).
+what to code, check out the following.
 
+### Learn More
+
+- We have [plenty of tutorials](https://substrate.dev/en/tutorials) to showcase
+Substrate development concepts and techniques.
+- To learn more about writing your own runtime with a front end, we have a
+[Substrate Collectables Workshop](https://substrate.dev/substrate-collectables-workshop)
+for building an end-to-end application.
+- For more information about runtime development tips and patterns, refer to our
+[Substrate Recipes](https://substrate.dev/recipes/).
+
+### Examples
+
+- Tutorial on [adding a `Contracts` module into runtime](tutorials/adding-a-module-to-your-runtime),
+so your node can run smart contract.
+
+### References
+
+- [The Cargo book](https://doc.rust-lang.org/stable/cargo/)
+- More about [Rust and WebAssembly](https://rustwasm.github.io/)
--- a/docs/tutorials/creating-a-runtime-module.md
+++ b/docs/tutorials/creating-a-runtime-module.md
@@ -2,46 +2,69 @@
 title: "Creating a Runtime Module"
 ---
 
-In this tutorial, you'll write a Substrate runtime module that lives in its own crate, and include it in a node based on the node-template.
+In this tutorial, you'll write a Substrate runtime module that lives in its own
+crate, and include it in a node based on the node-template.
 
 ## Setup Your Development Environment
 
-If you haven't already, follow the [fast installation](getting-started/installing-substrate.md#fast-installation) for getting started on Substrate. It will provide you with all the development resources you need to build Substrate and the [`substrate-up` scripts](getting-started/using-the-substrate-scripts) for quickly setting up new nodes.
+If you haven't already, follow the
+[fast installation](getting-started/installing-substrate#fast-installation) for
+getting started on Substrate. It will provide you with all the development
+resources you need to build Substrate and the
+[`substrate-up` scripts](getting-started/using-the-substrate-scripts) for
+quickly setting up new nodes.
 
-## Create your Node
+### Cloning Node and Module Template to Start
 
-We'll begin by creating a brand new node template. If you've followed other runtime tutorials, you've probably started this way before:
-
-```bash
-substrate-node-new modular-chain <your name>
-```
-
-In this tutorial, we're not going to write our module directly as part of the node template, but rather as a separate Rust crate. This approach provides a few advantages:
+We're not going to write our module directly as part of the node template, but
+rather as a separate Rust crate. This approach provides a few advantages:
 
 1. Our module can be easily imported into other nodes in the future.
-2. Our module can be published to places like GitHub or Crates.io without publishing our entire node in the same place.
+2. Our module can be published to places like GitHub or Crates.io without
+publishing our entire node in the same place.
 
-However, we do still need to set up a Substrate node to actually run our module.
+However, to effectively develop and test with your own runtime module, you need
+to first setup a single-node Substrate blockchain. Then inside the node runtime
+add in your own additional module.
 
-## Clone the Module Template Repository
-
-So if our module isn't going to live in the node template, where _is_ it going to live?
-
-If you aren't familiar with creating Rust libraries or using them in other Rust projects, you can get a more in-depth explanation by reading [the Cargo book](https://doc.rust-lang.org/cargo/guide/creating-a-new-project.html).
-
-While it's perfectly possible to create a new Substrate runtime module from scratch, the fastest way to get started is to clone a Substrate module template we have created for you:
+We'll begin from cloning the necessary repositories:
 
 ```bash
-git clone https://github.com/shawntabrizi/substrate-module-template
+cd my-playground-folder
+# cloning from a standard node template
+git clone https://github.com/substrate-developer-hub/substrate-node-template.git my-node
+# cloning from a standard module template
+git clone https://github.com/substrate-developer-hub/substrate-module-template.git my-module
+
+# kicking off the node compilation
+cd my-node
+cargo build --release
 ```
 
-In the `Cargo.toml` file, update the module's name and authorship to something meaningful. In this tutorial we're focusing on how to create and use the module rather than writing interesting module logic, so I'll call mine `test_module`. The beginning of my `Cargo.toml` now looks like.
+The compilation of the node may take up to 30 minutes, depending on your hardware.
+So let us kickstarted it first.
+
+Now, if our module isn't going to live in the node template, where _is_ it going
+to live? If you aren't familiar with creating Rust libraries or using them in
+other Rust projects, you can get a more in-depth explanation by reading
+[the Cargo book](https://doc.rust-lang.org/cargo/guide/creating-a-new-project.html).
+
+While it's perfectly possible to create a new Substrate runtime module from
+scratch, the fastest way is to clone from a Substrate module template, as we
+have just done above.
+
+In the `Cargo.toml` file, update the module's name and authorship to something
+meaningful. In this tutorial we're focusing on how to create and use the module
+rather than writing interesting module logic, so let's call it `test_module`.
+The beginning of the `Cargo.toml` now looks like:
+
+**`my-module/Cargo.toml`**
 
 ```toml
 [package]
 name = "test_module"
 version = "0.1.0"
-authors = ["Joshy Orndorff"]
+authors = ["Your Name"]
 edition = "2018"
 ```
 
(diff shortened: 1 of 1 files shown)


jimmychu0807:
thought: will be good if I can deploy the latest doc site and let the team review it. Probably write a guide so in future after we make an update, we can deploy the site in github under our own user namespace. (Or using heroku, etc..)

riusricardo:
> thought: will be good if I can deploy the latest doc site and let the team review it. Probably write a guide so in future after we make an update, we can deploy the site in github under our own user namespace. (Or using heroku, etc..)

I second that, we need a development environment to prevent working on production docs. For now this changes are being deployed after merging them into master and this has become our dev environment.

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -2,46 +2,68 @@
...
+crate, and include it in a node based on the node-template.
 
 ## Setup Your Development Environment
 
-If you haven't already, follow the [fast installation](getting-started/installing-substrate.md#fast-installation) for getting started on Substrate. It will provide you with all the development resources you need to build Substrate and the [`substrate-up` scripts](getting-started/using-the-substrate-scripts) for quickly setting up new nodes.
+If you haven't already, follow the
+[fast installation](getting-started/installing-substrate#fast-installation) for
+getting started on Substrate. It will provide you with all the development
+resources you need to build Substrate and the
+[`substrate-up` scripts](getting-started/using-the-substrate-scripts) for
```
shawntabrizi: You don't need to mention substrate-up scripts.

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -2,46 +2,68 @@
...
 
-In this tutorial, you'll write a Substrate runtime module that lives in its own crate, and include it in a node based on the node-template.
+In this tutorial, you'll write a Substrate runtime module that lives in its own
+crate, and include it in a node based on the node-template.
 
 ## Setup Your Development Environment
 
-If you haven't already, follow the [fast installation](getting-started/installing-substrate.md#fast-installation) for getting started on Substrate. It will provide you with all the development resources you need to build Substrate and the [`substrate-up` scripts](getting-started/using-the-substrate-scripts) for quickly setting up new nodes.
+If you haven't already, follow the
+[fast installation](getting-started/installing-substrate#fast-installation) for
```
shawntabrizi: We have a new getting-started page which doesnt have an "installing-substrate" sub page.

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -2,46 +2,68 @@
...
+getting started on Substrate. It will provide you with all the development
+resources you need to build Substrate and the
+[`substrate-up` scripts](getting-started/using-the-substrate-scripts) for
+quickly setting up new nodes.
 
-## Create your Node
+### Cloning Node and Module Template to Start
```
shawntabrizi:
```suggestion
### Cloning Node and Module Template
```

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -2,46 +2,68 @@
...
-substrate-node-new modular-chain <your name>
-```
-
-In this tutorial, we're not going to write our module directly as part of the node template, but rather as a separate Rust crate. This approach provides a few advantages:
+We're not going to write our module directly as part of the node template, but
+rather as a separate Rust crate. This approach provides a few advantages:
 
 1. Our module can be easily imported into other nodes in the future.
-2. Our module can be published to places like GitHub or Crates.io without publishing our entire node in the same place.
+2. Our module can be published to places like GitHub or Crates.io without
```
shawntabrizi: I learned pretty recently that we actually cannot upload this crate to `crates.io` because it has github dependencies, so we should probably remove this line.

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -2,46 +2,68 @@
...
+However, to effectively develop and test with your own runtime module, you need
+to first setup a single-node Substrate blockchain. Then inside the node runtime
+add in your own additional module.
 
-## Clone the Module Template Repository
+We'll begin from cloning the necessary repositories:
 
-So if our module isn't going to live in the node template, where _is_ it going to live?
+```bash
+cd my-playground-folder
```
shawntabrizi: this is a really small nitpick, but I think we should stick to simple folder names like: `example`

I could literally imagine the pain of users who have to type this in command line.

If it will be a "longer" name, then might as well be a descriptive one?

`runtime-module-tutorial`

shawntabrizi: Do you also need a `mkdir` here, or you assume the user knows to do that?

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -2,46 +2,68 @@
...
+# clone from a standard node template
+git clone https://github.com/substrate-developer-hub/substrate-node-template.git my-node
+# clone from a standard module template
+git clone https://github.com/substrate-developer-hub/substrate-module-template.git my-module
+# kick off the node compilation
+cd my-node
```
shawntabrizi: Is there a reason to change the name of the folder?

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -2,46 +2,68 @@
...
-However, we do still need to set up a Substrate node to actually run our module.
+However, to effectively develop and test with your own runtime module, you need
+to first setup a single-node Substrate blockchain. Then inside the node runtime
+add in your own additional module.
 
-## Clone the Module Template Repository
+We'll begin from cloning the necessary repositories:
 
-So if our module isn't going to live in the node template, where _is_ it going to live?
+```bash
```
shawntabrizi: You may want to break up this big code block into smaller ones since we enable copy and paste now of code blocks

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -2,46 +2,68 @@
...
+# clone from a standard module template
+git clone https://github.com/substrate-developer-hub/substrate-module-template.git my-module
+# kick off the node compilation
+cd my-node
+cargo build --release
+```
 
-If you aren't familiar with creating Rust libraries or using them in other Rust projects, you can get a more in-depth explanation by reading [the Cargo book](https://doc.rust-lang.org/cargo/guide/creating-a-new-project.html).
+The compilation of the node may take up to 30 minutes, depending on your hardware.
+So let us kickstarted it first.
```
shawntabrizi: This is not a valid English sentence.

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -2,46 +2,68 @@
...
+git clone https://github.com/substrate-developer-hub/substrate-node-template.git my-node
+# clone from a standard module template
+git clone https://github.com/substrate-developer-hub/substrate-module-template.git my-module
+# kick off the node compilation
+cd my-node
+cargo build --release
+```
 
-If you aren't familiar with creating Rust libraries or using them in other Rust projects, you can get a more in-depth explanation by reading [the Cargo book](https://doc.rust-lang.org/cargo/guide/creating-a-new-project.html).
+The compilation of the node may take up to 30 minutes, depending on your hardware.
```
shawntabrizi:
```suggestion
The compilation of the node may take up to 30 minutes depending on your hardware.
```

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -2,46 +2,68 @@
...
+So let us kickstarted it first.
 
-While it's perfectly possible to create a new Substrate runtime module from scratch, the fastest way to get started is to clone a Substrate module template we have created for you:
+Now, if our module isn't going to live in the node template, where _is_ it going
+to live? If you aren't familiar with creating Rust libraries or using them in
```
shawntabrizi: It would be good to clearly and directly answer the question you pose here, and then tell them to learn more from a link.

Otherwise, to me right now, it reads a bit like an open ended question.

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
 cargo build
```
shawntabrizi: Should we also use `--release` here for consistency?

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
 ```
 
-Let's explore some of the things we did for you in this Substrate module template. We will start by looking at the `Cargo.toml` file.
+Let's explore some of the things Substrate module template do for us, starting
+from the `Cargo.toml` file.
+
+## Development
```
shawntabrizi:
```suggestion
## Development

Let's explore some of the things Substrate module template do for us, starting
from the `Cargo.toml` file.
```

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
 ### Your Module's `std` Feature
 
-In your `Cargo.toml` file, you will notice a few lines about the "`std` feature". In Rust, when you enable `std`, you give your project access to [the Rust standard libraries](https://doc.rust-lang.org/std/). This works just fine when building native binaries.
+In your `my-module/Cargo.toml` file, you will notice a few lines about the
+"`std` feature". In Rust, when you enable `std`, you give your project access to
+[the Rust standard libraries](https://doc.rust-lang.org/std/). This works just
+fine when building native binaries.
 
-However, Substrate also builds the runtime code to Wasm. In this case we use cargo features to disable the Rust standard library when compiling for Wasm because [wasm-unknown-unknown](https://github.com/rust-lang/rust/pull/45905) does not know which allocator to use.
+However, Substrate also builds the runtime code to WebAssembly(Wasm). In this
```
shawntabrizi:
```suggestion
However, Substrate also builds the runtime code to WebAssembly (Wasm). In this
```

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
+case we use cargo features to disable the Rust standard library when compiling
+for Wasm because the compiler does not know which allocator to use. See
```
shawntabrizi: It would be good to get a strong follow up from a Wasm person the precise reasons here.

I dont think it i has to do with the allocator.

shawntabrizi: I thought it was all about binary size

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
+    'serde',
+    'codec/std',
     'support/std',
     'system/std',
+    'sr-primitives/std',
+    'runtime-io/std',
 ]
 ```
 
 > In general, when adding dependencies to your runtime, follow this pattern.
```
shawntabrizi:
```suggestion
> **Note:** In general, when adding dependencies to your runtime, follow this pattern.
```

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
+```TOML
+#--snip--
 
-# These dependencies specify a git branch
 [dependencies.support]
 default_features = false
 git = 'https://github.com/paritytech/substrate.git'
 package = 'srml-support'
-branch = 'v1.0'
+rev = '3dedd246c62255ba6f9b777ecba318dfc2078d85'
```
shawntabrizi: We should turn this to `branch = 'v2.0'` which will be our actual recommendation when this goes live.

Same applies for anywhere you used rev

shawntabrizi: Honestly, for a tutorial, it might be right just to tell them to use v2.0. Having multiple paths for the user is generally not a good idea.

We could add some notes or references to inform the user more if needed.

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
+
+[dependencies.support]
 default_features = false
 git = 'https://github.com/paritytech/substrate.git'
-package = 'srml-system'
-branch = 'v1.0'
-# Could also use a specific revision
-# rev = '<commit hash>'
+package = 'srml-support'
+tag = 'v1.0.0'
```
shawntabrizi: Why `tag` rather than `branch`?

AFAIK, `branch` may still get updates whereas a `tag` will not.

Imagine we find some really huge problem with the v1.0 branch, users who follow the tag method would not automatically get that update.

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
 
-The final section of the `Cargo.toml` file specifies the dev dependencies. These are the dependencies that are needed in your module's tests, but not necessary the actual module itself.
+The final section of the `Cargo.toml` file specifies the dev dependencies.
+These are the dependencies that are needed in your module's tests, but not
+necessary the actual module itself.
 
-
-```toml
+```TOML
+#--snip--
```
shawntabrizi: We should double check if we need a feature flag here or not. I think it was removed in the past, but done so incorrectly.

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
-You can confirm for yourself that the Substrate module template's test passes with:
+You can confirm for yourself that the Substrate module template's test passes
+with:
 
 ```bash
 cargo test
 ```
 
-> Optional: You can remove these dev dependencies from your `Cargo.toml` file and confirm that the module still compiles with `cargo build`. However, you will notice that `cargo test` does not compile! If you do this experiment, remember to restore the dev dependencies afterward.
+> Optional: You can remove these dev dependencies from your `Cargo.toml` file
```
shawntabrizi: IMO, remove this optional thing and just tell the user what will happen. Again, I think a tutorial should have a single well defined path.

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
+>
+> If you do this experiment, remember to restore the dev dependencies afterward.
+
+It will take a little experimentation to gain familiarity with what dependencies
+you need for each module you dream up. So the take-away is not that `Cargo.toml`
+should always look like it does in the template. Rather, `Cargo.toml` needs to
+have the correct dependencies and you now know how to specify them.
 
-It will take a little experimentation to gain familiarity with what dependencies you need for each module you dream up. So the take-away from this section is not that `Cargo.toml` should always look like it does in the template. Rather, `Cargo.toml` needs to have the correct dependencies and you now know how to specify them.
+### Add your Runtime Module to your Node
```
shawntabrizi:
```suggestion
### Add Your Runtime Module to Your Node
```

Use title case: http://titlecase.com/

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
+// add the following code block
 impl test_module::Trait for Runtime {
 	type Event = Event;
 }
-```
 
-3. Update `construct_runtime`
-```rust
-TestModule: test_module::{Module, Call, Storage, Event<T>},
+//...
```
shawntabrizi: Snip?

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
+    //...
+    // add the following line
+    TestModule: test_module::{Module, Call, Storage, Event<T>},
+  }
+);
 ```
 
-## Appreciate your Work
-At this point you have the template module packaged up as it's own crate, and included in your `modular-chain` runtime.
+### Appreciate your Work 😊
```
shawntabrizi:
```suggestion
### Appreciate Your Work 😊
```

Title case: http://titlecase.com/

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
-Build the `modular-chain` node with
 ```bash
-./scripts/build.sh
+cd my-playground-folder/my-node
+# compile your node
 cargo build --release
+# purge any existing chain
+cargo run --release -- purge-chain --dev
+# start the node
+cargo run --release -- --dev
```
shawntabrizi: Let's use `./target/release/node-template --dev`

Also note you may want to split this up since we do support copy and paste of code blocks now.

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
 ```
 
-And run your node with
-```bash
-# Clear the data directory in case you've done this before
-./target/release/modular-chain purge-chain --dev
+Finally, start the [Polkadot apps UI](https://polkadot.js.org/apps/#/explorer)
```
shawntabrizi:
```suggestion
Finally, start the [Polkadot-JS Apps](https://polkadot.js.org/apps/#/explorer)
```

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
+to confirm that the module is working as expected. In the apps UI, you will need
```
shawntabrizi: You could change the url to `https://polkadot.js.org/apps/#/explorer?rpc=ws://127.0.0.1:9944`

If you want to avoid telling these instructions.

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -51,184 +73,286 @@ You should be able to successfully compile this template module with:
...
+set to `Local Node`.
 
-# Start the chain
-./target/release/modular-chain --dev
-```
+### Publish your Module
 
-Finally, start the [Apps UI](https://polkadot.js.org/apps/#/explorer) to confirm that the module is working as expected. If you aren't familiar with the Apps UI, you will need to navigate to the settings tab, and select 'Local Node'.
+Once your module is no longer just testing code, you should consider publishing
+it. We'll cover publishing your module to GitHub here, but [crates.io](https://crates.io/)
```
shawntabrizi: As mentioned, I do not think you can use crates.io

shawntabrizi: @stiiifff confirm?

docs/tutorials/creating-a-runtime-module.md
JoshOrndorff: When I originally wrote this I didn't use `--release` because it takes longer and the build will never actually be executed. The module will be rebuilt when the entire node is built. Since writing this, I learned that `cargo check` is even faster than `cargo build` because it does all the syntax, type, and borrow checking but doesn't actually build an executable. I'd suggest we use `cargo check` here.

shawntabrizi: I am pretty happy with `cargo check`

I think, for docs purposes, if we ever do `cargo build` it should include `--release`, even if it is a little slower.

If they know rust well enough, they can omit the `--release` part, if they don't know rust well enough, better not have them later think that `cargo build` is okay for running a node.

jimmychu0807: why `cargo build` is not okay for running a node? My understanding is we get a "development version" of node, maybe more tracing, more verbose error/warning msg, less optimized memory usage, etc.

shawntabrizi: @jimmychu0807 yes, and those things can make it so that block production is stalled.

For example, I don't think you can upload smart contracts using a non-release build of a Substrate node. The transaction will always take longer than the 6-10 second blocktime, so then a block is not produced.

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -2,233 +2,318 @@
...
 
-[dependencies.system]
-default_features = false
-git = 'https://github.com/paritytech/substrate.git'
-package = 'srml-system'
-branch = 'v1.0'
-# Could also use a specific revision
-# rev = '<commit hash>'
+# To develop against a specific git commit, use:
+# rev = '3dedd246c62255ba6f9b777ecba318dfc2078d85'
```
riusricardo:
```suggestion
# rev = "<some commit hash>"
```

PR #296, opened 2019-10-23
Conceptual Doc: Substrate Keys

shawntabrizi:
Conceptual doc for session keys.

@joepetrowski is there more to expand on this page?

Things I would like to see expanded on:

- [x]  The different kinds of keys, where session keys are in the hierarchy of usage ("hotness"), and why we designed it this way.

- [x]  Information about application crypto for strong types across session keys

It could be that this page should be named "keys" and cover all the key types, not just session keys.


docs/conceptual/cryptography/keys.md
```diff
@@ -0,0 +1,145 @@
...
+* [Account Keys](#account-keys): Accounts that should hold funds.
+    * [Stash Keys](#stash-keys): Cold wallets controlled by controller keys.
+    * [Controller Keys](#controller-keys): Semi-online keys that will be in the
+      direct control of a user and used to submit extrinsics.
+* [Session Keys](#session-keys): Hot keys that be must kept online by a
+  validator to perform network operations and should not hold funds.
+
+## Account Keys
+
+Account keys represent the normal kind of accounts you might expect from other
```
joepetrowski: This is a bit redundant from the overview, but it is a good place to state clearly that keys != accounts.

------------------------

A key pair can represent an account and control funds, like normal accounts that you would expect from other blockchains. In the context of Substrate's Balances module [link], these accounts must have a minimum amount (an "existential deposit") to exist in storage.

Account keys are defined generically and made concrete in the runtime.

To continue with our example of Stash and Controller accounts, the keys to these accounts are distinguished by their intended use, not by any underlying cryptographic difference. When creating Stash or Controller keys, all cryptography supported for normal account keys are an option.

docs/conceptual/cryptography/keys.md
```diff
@@ -0,0 +1,145 @@
...
+1. Stash Keys
+2. Controller Keys
+
+These keys are distinguished by their intended use, not by any underlying
+cryptographic difference. When creating new controller or stash keys, all
+cryptography supported by account keys are an available option.
+
+### Stash Keys
+
+Stash keys are intended to behave like a cold wallet, existing on a piece of
```
joepetrowski: Should not mention voting here because you can actually designate a separate Proxy account. Controller can't vote unless you designate the same account as Proxy and Controller.

--------------------------

The Stash keys are the public/private key pair that defines a Stash account. This account is like a "savings account" in that you should not make frequent transactions from it. Therefore, its private key should be treated with the utmost security, for example protected in a safe or layers of hardware security.

Since the Stash key is kept offline, it designates a Controller account to make non-spending decisions with the weight of the Stash account's funds. It can also designate a Proxy account to vote in governance [link] on its behalf.

docs/conceptual/cryptography/keys.md
```diff
@@ -0,0 +1,145 @@
...
+at a bank, which ideally is only ever touched in urgent conditions.
+
+Since the stash key is kept offline, it must be set to have its funds bonded to
+a controller account. For non-spending actions, the controller has the funds of
+the stash behind it. For example, in nominating, staking, or voting, the
+controller can indicate its preference with the weight of the stash.
+
+### Controller Keys
+
+The controller key is a semi-online key that will be in the direct control of a
```
joepetrowski: The Controller keys are the public/private key pair that defines a Controller account. In the context of Substrate's NPoS model, the Controller key will signal one's intent to validate or nominate.

The Controller key is used to set preferences like rewards destination and, in the case of validators, to set their Session keys. The Controller account only needs to pay transaction fees, so needs a minimal amount of funds.

The Controller key can never be used to spend funds from its Stash account. However, actions taken by the Controller can result in slashing, so it should still be well secured.

docs/conceptual/cryptography/keys.md
```diff
@@ -0,0 +1,145 @@
...
+A controller key will never be able to move or claim the funds controlled the
+stash key. However, actions taken by the controller account can cause the stash
+funds to be slashed. So you still need to protect it and change it regularly.
+
+## Session Keys
+
+Session keys are "hot keys" that are used by validators to sign
+consensus-related messages. They are not meant to be used as account keys that
+control funds and should only be used for their intended purpose. They can be
+changed regularly; your controller only needs to create a certificate by signing
```
joepetrowski:
```suggestion
changed regularly; your Controller only needs to create a certificate by signing
```

docs/conceptual/cryptography/keys.md
```diff
@@ -0,0 +1,145 @@
...
+chain that this key represents their Controller key by publishing the Session
+certificate in a transaction on the chain.
+
+Substrate provides the [Session
+module](/rustdocs/master/srml_session/index.html), which allows validators to
+manage their session keys.
+
+### Strongly Typed Wrappers
+
+You can declare any number of session keys. For example, the default Substrate
```
joepetrowski:
```suggestion
You can declare any number of Session keys. For example, the default Substrate
```

docs/conceptual/cryptography/keys.md
```diff
@@ -0,0 +1,145 @@
...
+node uses three for BABE, GRANDPA, and "I'm Online". Other chains could have
+more or fewer depending on what operations the chain expects its validators to
+perform.
+
+These different session keys could use the same cryptography, but serve very
```
joepetrowski:
```suggestion
These different Session keys could use the same cryptography, but serve very
```

docs/conceptual/cryptography/keys.md
```diff
@@ -0,0 +1,145 @@
...
+different purposes throughout your runtime logic. To prevent the wrong key being
+used for the wrong operation, strong Rust types wrap these keys, keeping them
+incompatible with one another and ensuring they are only used for their intended
+purpose.
+
+### Generation
+
+If a Session keys is compromised, attackers could commit slashable behavior.
```
joepetrowski:
```suggestion
If a Session key is compromised, attackers could commit slashable behavior.
```

shawntabrizi:
Thanks for the suggestions @joepetrowski I have merged them in.

I still need to work on the Crypto page, it is far from the right spot. Just copy and paste so far

docs/conceptual/cryptography/index.md
```diff
@@ -0,0 +1,44 @@
+---
+title: Cryptography
+---
+
+This is a high-level overview of the cryptography used in Polkadot.
```
JoshOrndorff: Why are we describing specifics of Polkadot in devhub?

riusricardo: Some of the text for this PR was taken from the Polkadot wiki. It could be better to generalize it.

shawntabrizi: Yes, right now it is just a copy paste from the polkadot page. I need to redo the whole page.

I might save it for another PR.

docs/conceptual/cryptography/index.md
```diff
@@ -0,0 +1,97 @@
...
+
+[SR25519](https://research.web3.foundation/en/latest/polkadot/keys/1-accounts-more/) is based on the
+same underlying curve as [Ed25519](#ed25519). However, it uses Schnorr signatures instead of the
+EdDSA scheme.
+
+Schnorr signatures bring some noticeable features over the [ECDSA](#ecdsa)/EdDSA schemes:
+
+* It is better for hierarchical deterministic key derivations.
+
+* It allows for native multi-signature through [signature
```
riusricardo: This is true but we can't do it right now on Substrate. Should we add a `Note` to make it clear that it is WIP?

shawntabrizi: I think it is doable and doable outside the context of Substrate. Like we could generate multi-signature keys which are compatible with Substrate's account system and represented as a single account on chain, but we have some multi-sig logic off chain.

I would want to confirm what I think is right.

shawntabrizi:
> I read the content. Content looks good to me, but I have no knowledge to verify what is said is true or not.
> 
> Shawn, I would think you need the specific implementer to proof-read for correctness?

Yeah. @burdges should help give a final update on content.

docs/conceptual/cryptography/index.md
```diff
@@ -0,0 +1,97 @@
...
+  aggregation](https://bitcoincore.org/en/2017/03/23/schnorr-signature-aggregation/).
+
+* It is generally more resistant to misuse.
```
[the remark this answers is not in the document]
shawntabrizi: misuse is right here for English.

docs/conceptual/cryptography/index.md
```diff
@@ -0,0 +1,97 @@
...
+xxHash is used in places where outside parties cannot manipulate the input of the hash function. For
+example, it is used to generate the key for runtime storage values, whose inputs are controlled by
+the runtime developer.
+
+Substrate uses the [`twox-hash`](https://github.com/shepmaster/twox-hash) implementation in Rust.
+
+### Blake2
+
+[Blake2](https://en.wikipedia.org/wiki/BLAKE_(hash_function)#BLAKE2) is a cryptographic hash
+function. It is considered to be a very fast and is also used in [Zcash](https://z.cash).
```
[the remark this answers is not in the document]
shawntabrizi: Updated the link to Wikipedia, good suggestion

docs/conceptual/cryptography/index.md
```diff
@@ -0,0 +1,97 @@
...
+Substrate provides multiple different cryptographic schemes and is generic such that it can support
+anything which implements the [`Pair`
+trait](https://substrate.dev/rustdocs/master/substrate_primitives/crypto/trait.Pair.html).
+
+### ECDSA
+
+Substrate provides an
+[ECDSA](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm) signature scheme
+using the [secp256k1](https://en.bitcoin.it/wiki/Secp256k1) curve. This is the same cryptographic
+algorithm used to secure [Bitcoin](https://en.wikipedia.org/wiki/Bitcoin) and
```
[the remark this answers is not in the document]
shawntabrizi: I think reasonable reference docs can help people conceptually understand Substrate. We make assumptions that the user is quite informed of blockchains, and these links can help someone who is not. imo.

PR #305, opened 2019-11-08
Update Home Page

shawntabrizi:
* Reduces padding on hero section
* Changes the call to actions to become "who" page:
    * Contract
    * Runtime
    * Front-end
* Updates getting started link to the "substrate install" page

image

### DIFF
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -90,7 +90,7 @@ class Index extends React.Component {
             href={pageUrl('who/runtime-developer')}
             className='m-1 primary-color'
           >
-            <translate>Start Here</translate>
+            <translate>Get Started on Substrate</translate>
           </Button>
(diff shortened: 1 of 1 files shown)


shawntabrizi:
Need to update the names of the buttons. Can't be "Get Started", and then three "Start Here".

meheszsofi:
do we need the "get started" button on the top at all? without it there are 3 very straightforward paths of navigation from the homepage. 
because it is on the top banner in the middle now the "get started" button is the strongest in the hierarchy, if we want people to rather choose between 3 others (runtime/frontend/smart contract) it might make sense to simply get rid of it (or maybe rename it and put it in a less central position)

shawntabrizi:
@meheszsofi we probably don't, can update.

shawntabrizi:
image

Updated

PR #311, opened 2019-11-10
Substrate Overview Page

shawntabrizi:


docs/index.md
```diff
@@ -2,4 +2,74 @@
...
+
+- View our developer documentation for [using the SRML for module
+  development](development/module/index.md).
+
+### Examples
+
+- Follow our [tutorial to create your first custom blockchain with
+  Substrate](tutorials/creating-your-first-substrate-chain/index.md).
+
+- Follow our [tutorial to start a private network with
```
kaichaosun: How about we put this one as the first example? Since it doesn't require any programming and user can feel substrate.

docs/index.md
```diff
@@ -2,4 +2,74 @@
...
+
+Despite being "completely generic", it comes with both standards and conventions - particularly with
+the Substrate Runtime Module Library ([SRML](conceptual/runtime/srml.md)) - regarding the underlying
+data-structures that power the STF, thereby making rapid blockchain development a reality.
+
+## Usage
+
+![Technical Freedom vs Development Ease](assets/technical-freedom.png)
+
+Substrate is designed to be used in one of three ways:
```
kaichaosun: Our users usually start with substrate node template to write their modules or add more srml modules, so that finally start a new chain. In this overview page, should we document such lifecycle? Here you are trying to tell user the three ways that they can play with substrate, but seems two of them (1, 3) are not very common usage.

docs/index.md
```diff
@@ -2,4 +2,74 @@
 title: Overview
 ---
 
-My new content here..
\ No newline at end of file
+This page is the top level entry point into the Substrate documentation.
+
+> **Some Expertise Needed**
+>
+>In order to get the most out of Substrate, you should have a good knowledge of blockchain concepts
```
kaichaosun:
```suggestion
>In order to get the most out of Substrate, you should have a basic knowledge of blockchain concepts
```

docs/index.md
```diff
@@ -2,4 +2,74 @@
...
+>and basic cryptography. Terminology like header, block, client, hash, transaction and signature
```
kaichaosun:
```suggestion
>. Terminology like header, block, client, hash, transaction and signature
```

docs/index.md
shawntabrizi: This is messaging coming from Gav. At this point in our documentation, we are not leading users to do a certain thing, just telling them what substrate can do. Creating modules with the SRML is a specific way to approach using Substrate.

riusricardo: Fair point, the Substrate binary is only used in the ink! tutorial. However it uses all the modules and it is a good example. I think that it is correct to keep it high level. To me, this page is about showing Substrate features and not getting into details.

PR #318, opened 2019-11-11
Sub0 Blast

shawntabrizi:
Updates blast logic.

- Makes it a footer instead of a header, fixing all spacing issues
- Adds padding to bottom
- Allows image instead of text for content

image


shawntabrizi:
Update:
image

PR #319, opened 2019-11-11
Update Dark Color

shawntabrizi:
Merge after https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/318

Maybe not the cleanest between the blast and the bottom

image


JoshOrndorff:
@shawntabrizi Closed because the dark theme is only for sub0 right now right?

Or was it something else. FWIW I do like the way it looks, but I agree to close it if design team prefers not to use those colors yet.

shawntabrizi:
@JoshOrndorff Closed because I believe we agreed to do some bigger CSS overhaul with them when we are back in Berlin

Issue #327, opened 2019-11-15
suggestion what to add to 'start-a-private-network-with-substrate'


drandreaskrueger:
see this comment here: https://github.com/paritytech/substrate/issues/4096#issuecomment-553936224

the 3rd paragraph "I think this page needs some modifications/additions ..."

shawntabrizi:
@JoshOrndorff This can be closed?

PR #329, opened 2019-11-15
Revise private network tutorial

JoshOrndorff:
This PR contains several changes and revisions to the private network tutorial:
* Pagination
* Use node template as starting point
* Multiple key generation options
* Details recommended in #327


shawntabrizi:
@JoshOrndorff Add some next steps please :)

docs/tutorials/start-a-private-network/1compile.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Compiling the Tools
+---
+
+Substrate does not (yet) offer binary installation packages, so it must be compiled from source, which can be a time-consuming process. It is wise to being the compilation, and then pre-read the next few pages of the tutorial while your node is building. This tutorial will use the Substrate node template.
```
shawntabrizi:
```suggestion
Substrate does not (yet) offer binary installation packages, so it must be compiled from source, which can be a time-consuming process. It is wise to start the compilation and then read the next few pages of the tutorial while your node is building. This tutorial will use the Substrate node template.
```

docs/tutorials/start-a-private-network/2alicebob
```diff
@@ -0,0 +1,106 @@
...
+2019-09-26 15:42:10 Using default protocol ID "sup" because none is configured in the chain specs
+2019-09-26 15:42:10 Local node identity is: QmViTYm8Yr4yneyRGov9L87vLeVLoySFduVRNQ1d33WdXN
+2019-09-26 15:42:10 Starting BABE Authorship worker
+2019-09-26 15:42:15 Idle (0 peers), best: #0 (0x8855…e564), finalized #0 (0x8855…e564), ⬇ 0 ⬆ 0
+2019-09-26 15:42:20 Idle (0 peers), best: #0 (0x8855…e564), finalized #0 (0x8855…e564), ⬇ 0 ⬆ 0
+```
+
+A few lines to take note of:
+`Local node identity is: QmViTYm8Yr4yneyRGov9L87vLeVLoySFduVRNQ1d33WdXN` shows the node ID that Bob will need when booting from Alice's node.
+`Initializing Genesis block/state (state: 0x4faf…3aba, header-hash: 0x8855…e564)` tells which genesis block the node is using. Bob's node must have the same numbers or they will not connect to one another.
```
shawntabrizi:
```suggestion
`Initializing Genesis block/state (state: 0x4faf…3aba, header-hash: 0x8855…e564)` tells which genesis block the node is using. Bob's node must have the same hashes or they will not connect to one another.
```

docs/tutorials/start-a-private-network/2alicebob
```diff
@@ -0,0 +1,106 @@
...
+
+More details about all of these flags and others that I haven't mentioned are available by running `./target/release/node-template --help`.
+
+## Attach a UI
+You can tell a lot about your node by watching the output it produces in your terminal. There is also a nice graphical user interface known as the [Polkadot Js Apps UI](https://polkadot.js.org/apps/) which you can connect to your node. That link goes to the hosted version of the user interface, which is super convenient, but will not connect to unencrypted web-socket endpoints in some browsers (notably including Firefox). In this case, you can run the interface locally by grabbing the code from [github](https://github.com/polkadot-js/apps). In general the instructions in that repository will be your best guide, but the process should be something like this.
```
shawntabrizi:
```suggestion
You can tell a lot about your node by watching the output it produces in your terminal. There is also a nice graphical user interface known as the [Polkadot-JS Apps UI](https://polkadot.js.org/apps/) which you can connect to your node. That link goes to the hosted version of the user interface, which is super convenient, but will not connect to unencrypted web-socket endpoints in some browsers (notably including Firefox). In this case, you can run the interface locally by grabbing the code from [github](https://github.com/polkadot-js/apps). In general the instructions in that repository will be your best guide, but the process should be something like this.
```

docs/tutorials/start-a-private-network/3keygen.md
```diff
@@ -0,0 +1,125 @@
+---
+title: Generate Your Own Keys
+---
+
+Now that we know the fundamentals and the command line options, it's time to generate our own keys rather than using the well-known Alice and Bob keys. Each person who wants to participate in the blockchain should generate their own keys. This page explains several options for generating keys, and each participant only needs to choose one. Regardless of which option you choose, be sure to record all of the output from this section as you will need it later.
+
+## Option 1: Subkey
+Subkey is a tool the generates keys specifically designed to be used with Substrate. This is the most secure way to generate keys, and is how you should generate keys if you plan to use them in production.
+
+Begin by compiling and install the utility.
```
shawntabrizi:
```suggestion
Begin by compiling and installing the utility.
```

docs/tutorials/start-a-private-network/3keygen.md
```diff
@@ -0,0 +1,125 @@
+---
+title: Generate Your Own Keys
+---
+
+Now that we know the fundamentals and the command line options, it's time to generate our own keys rather than using the well-known Alice and Bob keys. Each person who wants to participate in the blockchain should generate their own keys. This page explains several options for generating keys, and each participant only needs to choose one. Regardless of which option you choose, be sure to record all of the output from this section as you will need it later.
+
+## Option 1: Subkey
+Subkey is a tool the generates keys specifically designed to be used with Substrate. This is the most secure way to generate keys, and is how you should generate keys if you plan to use them in production.
```
shawntabrizi: I think you should omit "this is the most secure... how you should generate keys".

We can create a separate page on securely generating keys if you would like to reference something like it, but I don't think this is advice we want to have on our docs.

The best and most secure way to generate keys is offline on a device which has never been on or will be on the internet.

```suggestion
Subkey is a tool the generates keys specifically designed to be used with Substrate.
```

docs/tutorials/start-a-private-network/3keygen.md
shawntabrizi: I might also suggest that you be honest with the user that this might take a while to compile. Maybe 15 min or more.

docs/tutorials/start-a-private-network/4customchain.md
```diff
@@ -0,0 +1,115 @@
+---
+title: Creating Your Private Network
+---
+Now that each participant has their own keys generated, you're ready to start your own custom chain.
+
+> Validators should not share the same keys, even for learning purposes. If two validators have the same keys, they will produce conflicting blocks, and be slashed.
```
shawntabrizi:
```suggestion
> Validators should not share the same keys, even for learning purposes. If two validators have the same keys, they will produce conflicting blocks and be slashed.
```

docs/tutorials/start-a-private-network/4customchain.md
```diff
@@ -0,0 +1,115 @@
...
+
+Last time around, we used `--chain=local` which is a predefined "chainspec" that has Alice and Bob specified as validators along with many other useful defaults.
+
+Rather than writing our chainspec completely from scratch, we'll just make a few modifications to the the one we used before. To start we need to export the chainspec to a json file. Remember, further details about all of these commands are available by running `node-template --help`.
+
+```bash
+./target/release/node-template build-spec --chain=local > customSpec.json
+```
+
+The file we just created contains several fields, and one can learn a lot by exploring them. By far the largest field is a single hex number that is the Wasm binary of our runtime. It is part of what you built earlier when you ran the `cargo build` command. Having this information on-chain enables forkless upgrades.
```
shawntabrizi:
```suggestion
The file we just created contains several fields, and you can learn a lot by exploring them. By far the largest field is a single hex number that is the Wasm binary of our runtime. It is part of what you built earlier when you ran the `cargo build` command. Having this information on-chain enables forkless upgrades.
```

docs/tutorials/start-a-private-network/4customchain.md
```diff
@@ -0,0 +1,115 @@
...
+    [
+      "5GoNkf6WdbxCFnPdAnYYQyCjAKPJgLNxXwPjwTh6DGg6gN3E",
+      1
+    ]
+  ]
+},
+
+```
+
+All we need to do is change the authority addresses listed (currently Alice and Bob) to our own addresses that we generated in the previous step. The sr25519 addresses go in the `aura` section, and the ed25519 addresses in the `grandpa` section. You may add as many validators as you like. For additional context, you can see how polkadot uses these keys and others in the [Polkadot wiki](https://wiki.polkadot.network/docs/en/learn-keys#session-keys).
```
shawntabrizi: Can we use our own docs here? I believe we have similar data about session keys.

shawntabrizi: https://substrate.dev/docs/en/next/conceptual/cryptography/keys

docs/tutorials/start-a-private-network/4customchain.md
```diff
@@ -0,0 +1,115 @@
...
+
+> A single person should do these steps and share the resulting file with their fellow validators.
```
shawntabrizi:
```suggestion
> **Note:** A single person should do these steps and share the resulting file with their fellow validators.
```

docs/tutorials/start-a-private-network/4customchain.md
```diff
@@ -0,0 +1,115 @@
...
+>
+> Because Rust -> Wasm builds aren't "reproducible", each person will get a slightly different Wasm blob which will break consensus if each participant generates the file themselves.
+
+Once the chainspec is prepared, convert it to a "raw" chainspec. The raw chainspec contains all the same information, but it contains the encoded storage keys that the node will use to reference the data in its local storage. Distributing a raw spec ensures that each node will store the data at the proper storage keys.
```
shawntabrizi: I think this will be confusing for someone new to Substrate.
```suggestion
Once the chainspec is prepared, convert it to a "raw" chainspec. The raw chainspec is pre-processed by your node to ensure that everyone who uses it gets the same final result in their genesis block.
```

docs/tutorials/start-a-private-network/4customchain.md
```diff
@@ -0,0 +1,115 @@
+---
+title: Creating Your Private Network
+---
+Now that each participant has their own keys generated, you're ready to start your own custom chain.
+
+> Validators should not share the same keys, even for learning purposes. If two validators have the same keys, they will produce conflicting blocks, and be slashed.
+
+## Create a Chainspec
```
shawntabrizi:
```suggestion
## Create a Chain Specification
```

I think it is "chain spec" rather than one word. Would recommend doing a find replace here.

docs/tutorials/start-a-private-network/4customchain.md
```diff
@@ -0,0 +1,115 @@
...
+
+Once the chainspec is prepared, convert it to a "raw" chainspec. The raw chainspec contains all the same information, but it contains the encoded storage keys that the node will use to reference the data in its local storage. Distributing a raw spec ensures that each node will store the data at the proper storage keys.
+
+```bash
+./target/release/node-template build-spec --chain customSpec.json --raw > customSpecRaw.json
+```
+
+Finally share the `customSpecRaw.json` with your all the other validators in the network.
+
+## First Participant starts a Bootnode
```
shawntabrizi:
```suggestion
## First Participant Starts a Bootnode
```

docs/tutorials/start-a-private-network/4customchain.md
```diff
@@ -0,0 +1,115 @@
...
+
+Finally, restart your node so that the keys just added to the keystore take effect.
+
+## Subsequent Participants Join
+
+Subsequent validators can now join the network as Bob did previously, making sure to use the new chainspec. You may bootstrap from _any_ of the nodes already in the network, not just the one that went first.
+
+Each participant will need to start their node, add both keys to the keystore, and restart their node.
+
+> If you are running multiple nodes on the same machine, you must direct the UI to the correct node's websocket endpoint.
```
shawntabrizi:
```suggestion
> **Note:** If you are running multiple nodes on the same machine, you must direct the UI to the correct node's websocket endpoint.
```

docs/tutorials/start-a-private-network/2alicebob
```diff
@@ -0,0 +1,106 @@
+---
+title: Alice and Bob Start Blockchain
+---
+
+Before we generate our own keys, and start a truly unique Substrate network, let's learn the fundamentals by starting with a pre-defined network specification called `local` with two pre-defined (and definitely not private!) keys known as Alice and Bob.
+
+## Alice Starts first
```
shawntabrizi:
```suggestion
## Alice Starts First
```

docs/tutorials/start-a-private-network/2alicebob
```diff
@@ -0,0 +1,106 @@
...
+
+Alice (or whomever is playing her) should run this command from node-template repository root.
+```bash
+# Start Alice'snode
```
shawntabrizi:
```suggestion
# Start Alice's node
```

docs/tutorials/start-a-private-network/4customchain.md
```diff
@@ -0,0 +1,115 @@
...
+  --chain ./customSpecRaw.json \
+  --port 30333 \
+  --ws-port 9944 \
+  --rpc-port 9933 \
+  --telemetry-url ws://telemetry.polkadot.io:1024 \
+  --validator
+  --name MyNode
+```
+Here are some differences from when we launched as Alice.
+* I've omitted the `--base-path` flag, so Substrate will use a default location. You're still free to specify your own base path, and you must do so if multiple nodes will run on the same physical machine.
```
shawntabrizi: The thought here is that the user would set up another local network with 2 machines? Why not have them specify a base path again? It seems they would have to.

In either case, I think it makes sense to teach them about `purge-chain` if you have been teaching them to delete the folder up until now.

docs/tutorials/start-a-private-network/3keygen.md
```diff
@@ -0,0 +1,125 @@
...
+Now see the `ed25519` key and address associated with the same mnemonic. This key will be used by GRANDPA for block finalization.
+```
+$ subkey --ed25519 inspect "keep matrix knee meat awake frown rubber position federal easily strategy inhale"
+Secret phrase `keep matrix knee meat awake frown rubber position federal easily strategy inhale` is account:
+  Secret seed: 0xb5d5cda89e139aecb67181e11d6d2d90a0cc80106afa035ab19264af7b5e5c0b
+  Public key (hex): 0xfe68fdff17960cb8d45d861396a64d4086997353849403ee3352996ec68ff4af
+  Address (SS58): 5HpHD5YseSWbHfni43Zm2SjtpyqSVmuaNhKkVmivp8L93Trs
+```
+
+## Option 2: Polkadot JS Apps
```
shawntabrizi:
```suggestion
## Option 2: Polkadot-JS Apps
```

docs/tutorials/start-a-private-network/3keygen.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Option 1: Subkey
+Subkey is a tool the generates keys specifically designed to be used with Substrate. This is the most secure way to generate keys, and is how you should generate keys if you plan to use them in production.
+
+Begin by compiling and install the utility.
+```bash
+cargo install --force subkey --git https://github.com/paritytech/substrate
+```
+
+Generate a mnemonic and see the `sr25519` key and address associated with it. This key will be used by Aura for block production.
```
kaichaosun: It's better if there's a link to Aura's detail description.

docs/tutorials/start-a-private-network/3keygen.md
```diff
@@ -0,0 +1,125 @@
...
+```
+$ subkey generate
+Secret phrase `keep matrix knee meat awake frown rubber position federal easily strategy inhale` is account:
+  Secret seed: 0xb5d5cda89e139aecb67181e11d6d2d90a0cc80106afa035ab19264af7b5e5c0b
+  Public key (hex): 0x8ed5f822065e5824d3e37d9ea36a81eacb98ff1a6fa04bb87d2fa4915e9ed147
+  Address (SS58): 5FHzDem7A5aAq79tuEN9xJuNPXiYfmRQamhumTuqu6i57BuU
+```
+
+Now see the `ed25519` key and address associated with the same mnemonic. This key will be used by GRANDPA for block finalization.
```
kaichaosun: Same with Aura, GRANDPA need a link.

docs/tutorials/start-a-private-network/4customchain.md
```diff
@@ -0,0 +1,115 @@
...
+* I've omitted the `--alice` flag. Instead we will insert our own custom keys into the keystore through the RPC shortly.
+* The `--chain` flag has changed to use our custom chainspec.
+* I've added the optional `--name` flag. You may use it to give your node a human-readable name in the telemetry UI.
+
+One your node is running, you will again notice that no blocks are being produced. At this point you can use the Apps UI to insert your keys into the keystore. Navigate to the "Toolbox" tab and the "RPC Call" sub-tab. Choose "author" and "insertKey". The fields can be filled like this:
```
kaichaosun:
```suggestion
Once your node is running, you will again notice that no blocks are being produced. At this point you can use the Apps UI to insert your keys into the keystore. Navigate to the "Toolbox" tab and the "RPC Call" sub-tab. Choose "author" and "insertKey". The fields can be filled like this:
```

docs/tutorials/start-a-private-network/4customchain.md
kaichaosun:
```suggestion
> If you are running multiple nodes on the same machine, you must configure the UI to the correct node's WebSocket endpoint.
```

docs/tutorials/start-a-private-network/1compile.md
```diff
@@ -0,0 +1,25 @@
+---
```
kaichaosun: I will be more happy if the filename is `1_compile.md`.

docs/tutorials/start-a-private-network/3keygen.md
kaichaosun: How about put Polkadot JS Apps as Option 1? Since it's more efficient for testing purpose.

kaichaosun:
Very comprehensive writeup, I like it. 👍

docs/tutorials/start-a-private-network/2alicebob
```diff
@@ -0,0 +1,106 @@
+---
+title: Alice and Bob Start Blockchain
```
kaichaosun: Is local dev network also useful in this tutorial (the network start with `--dev`)? As it's probably the first tutorial that the user interact with substrate.

docs/tutorials/start-a-private-network/3keygen.md
riusricardo: It may be slow but could be the best approach in the long term. As you said we may want to have a separate section for generating keys and link it when needed.
The UI could be changing faster than `subkey` and it adds another thing to keep up to date every time it changes.
On the other side it is easier to use the UI than `subkey` so I think that we need to find the correct balance here.

docs/tutorials/start-a-private-network/3keygen.md
riusricardo: I think that the statement is correct but it could be more info than what is needed here.

riusricardo: Thinking more about this statement. I think that is should be removed. The main reason is that even if it is true your console output is not encrypted and it is not the most secure way. It will depend on how the user stores the seed or private key.
In case of the UI there are also some attack vectors but at least the file that you get is encrypted and generated locally on client side.

docs/tutorials/start-a-private-network/3keygen.md
riusricardo: At the end it doesn't matter for this tutorial. You need to enter the `seed` for becoming a validator. It doesn't matter how you get it. But I think that because this is not a beginners tutorial, it should be okay to introduce a new tool.

docs/tutorials/start-a-private-network/2alicebob
JoshOrndorff: I don't get it. Did you change something?

docs/tutorials/start-a-private-network/4customchain.md
JoshOrndorff: I believe it is correct as written. https://www.grammarly.com/blog/comma-before-and/ has the following example 
image

Simplifying my sentence it becomes:
Incorrect: They will produce blocks, and be slashed.
  Correct: They will produce block and be slashed.

Even though nobody asked: Although it appears syntactically similar, this is not an example of the [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma) of which I'm a big fan.

docs/tutorials/start-a-private-network/4customchain.md
JoshOrndorff: one -> you is fine with me. Why do you prefer to remove the bit about forkless upgrades?

shawntabrizi: It is just really out of context. I dont remember you really making mention of it anywhere else, and then you drop it pretty casually like the reader knows what that means.

You def can elaborate on it, or better, just keep your tutorial concise about what you are trying to teach, and expand on forkless upgrades as a "learn more" step

docs/tutorials/start-a-private-network/4customchain.md
JoshOrndorff: Personally, I don't think the word "Note" adds anything. It wouldn't be in that highlighted box if you weren't supposed to note it.

docs/tutorials/start-a-private-network/4customchain.md
JoshOrndorff: I agree that a newcomer will not understand all the detail. But I think they'll understand, "It's better to use the raw one for some subtle reason". However a moderate uer may understand and appreciate the detail.

docs/tutorials/start-a-private-network/2alicebob
JoshOrndorff: This tutorial is about starting a non-trivial multi-node network. `--dev` is useful for developing dApps and users will encounter it in tutorials about developing dApps.

docs/tutorials/start-a-private-network/2alicebob
shawntabrizi: I changed "Polkadot-JS Apps" and "GitHub"

docs/tutorials/start-a-private-network/4customchain.md
shawntabrizi: I am suggesting to remove the comma, which you seem to agree is correct?

docs/tutorials/start-a-private-network/4customchain.md
shawntabrizi: I think you should strive for clarity, not just drop knowledge out of context. Also, I think we should be honest about the audience for this tutorial, which is likely someone who will not understand the detail.

docs/tutorials/start-a-private-network/4customchain.md
JoshOrndorff: Guess I forgot how to read diffs there for a sec.

docs/tutorials/start-a-private-network/3keygen.md
JoshOrndorff: I agree I'll draft a PR for those write-ups so we can link to the. I don't want that to block this PR though.

docs/tutorials/start-a-private-network/1compile.md
JoshOrndorff: Took shawns advice to remove them. I do still prefer to keep them, and would be happy to type-set them this way.

PR #334, opened 2019-11-19
Add sidebars for private network tutorial

JoshOrndorff:
In my recent PR to update the private network tutorial, I failed to add the proper sidebars to the tutorial. This solves that.

We will also need to update the link on from the tutorials index when we launch v2.0

### DIFF
--- a/website/sidebars.json
+++ b/website/sidebars.json
@@ -118,5 +118,14 @@
       "tutorials/creating-your-first-substrate-chain/build",
       "tutorials/creating-your-first-substrate-chain/front-end"
     ]
+  },
+  "start-a-private-network": {
+    "Start a Private Network": [
+      "tutorials/start-a-private-network/index",
+      "tutorials/start-a-private-network/compile",
+      "tutorials/start-a-private-network/alicebob",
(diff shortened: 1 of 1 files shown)


riusricardo:
Joshy I'm probably mistaken here, but does this PR is trying to fix an invalid link?
>What happened with this link? https://substrate.dev/docs/en/next/tutorials/start-a-private-network-with-substrate.md

It may be good enough to just do this for the sidebar and not get into more routing.

JoshOrndorff:
@riusricardo This PR does not fix the broken link you asked about. It merely makes it possible to navigate between the pages of the new tutorial with the sidebars.

That link is broken because I changed the name of the tutorial:
```diff
- Start a private network with Substrate
+ Start a private network
```

I like the new title better because everything on the devhub is about Substrate, so it doesn't need to be included in each title. The broken link is an unfortunate side-effect.

One solution would be to go back and change the v1.0 tutorial's title as well. Then the "add 'next'" trick would still work, but that specific link you quoted will never work again.

Another option is to just do nothing about it because we're switching to v2.0 docs soon.

PR #336, opened 2019-11-20
Cover Consensus Topics

JoshOrndorff:
This PR is currently a draft to add coverage of consensus topics to devhub.

I'd like some preliminary feedback about the scope, and structure of this article before I do the following.
- [ ] Write individual Articles for
  - [ ] Grandpa
  - [ ] Babe
  - [ ] Aura
  - [ ] PoW
- [ ] Fill in Diagrams, Links, other todos
- [ ] Solicit feedback from Rob, Andre, maybe others
- [ ] Add menu items


docs/conceptual/core/consensus.md
```diff
@@ -2,14 +2,104 @@
...
+### Voting Conflict
+A blockchain runtime tallys simple two-option votes for a  DAO. When a new election is opened, Alice, a member of the DAO, can vote Aye. Alice could also vote Nay. But Alice cannot vote both Aye and Nay in the same election.
 
-* Finality
+### Land Registry Conflict
+Consider a blockchain runtime that serves as a land registry for a futuristic nation-state. The registry keeps track of transfers of land ownership among citizens, and occasionally opens new plots of land to be claimed. For a particular newly opened land run, Alice could claim the plot, or Bob could claim it. But they cannot both claim it.
 
-* NPoS
+### Classic Double Spend Conflict
+A pure cryptocurrency, like Bitcoin, tracks the minting and spending of tokens. Alice holds a particular coin, so she can pay it to Bob. She could also pay that coin to Carol. But she cannot pay the same coin to bot Bob and Carol.
```
4meta5:
```suggestion
A pure cryptocurrency, like Bitcoin, tracks the minting and spending of tokens. Alice holds a particular coin, so she can pay it to Bob. She could also pay that coin to Carol. But she cannot pay the same coin to both Bob and Carol.
```

docs/conceptual/core/consensus.md
```diff
@@ -2,14 +2,104 @@
...
 
-* PoW
+The unifying property in these examples is that there are multiple transactions, each of which is individually valid, but are mutually exclusive. Among the alternatives, a single transaction must be chosen.
 
-* What would it take to write my own consensus?
+
+## Conflict Exclusion
+In centralized systems, the central authority chooses among mutually exclusive alternatives by recording transactions in the order it sees them, and choosing the first of the competing alternatives when a conflict arises. In decentralized systems, the nodes will see transaction is different orders, and thus they must use a more elaborate method to exclude transactions. As a further complication, blockchain networks strive to be fault tolerant, which means that they should continue to provide consistent data even if up to one third of participants are not following the rules.
```
4meta5:
```suggestion
In centralized systems, the central authority chooses among mutually exclusive alternatives by recording transactions in the order it sees them, and choosing the first of the competing alternatives when a conflict arises. In decentralized systems, the nodes will see transactions in different orders, and thus they must use a more elaborate method to exclude transactions. As a further complication, blockchain networks strive to be fault tolerant, which means that they should continue to provide consistent data even if up to one third of participants are not following the rules.
```

docs/conceptual/core/consensus.md
```diff
@@ -2,14 +2,104 @@
...
+
+One practical solution, and the one taken in Substrate, is to find a decentralized way to agree on transaction ordering, and then resolve conflicts by selecting the first alternative seen just as the centralized authority did.
+
+> In general a blockchain just needs some method to choose among conflicting transactions, and there are interesting paradigms that do not involve a total ordering. Such methods will not be considered further here.
+
+## Fork Choice Rules
+
+A block contains many transaction (technically it contains zero or more transactions), and a reference to its parent block. A blockchain is any block, and all of its parents recursively.
```
4meta5:
```suggestion
A block contains many transactions (technically it contains zero or more transactions), and a reference to its parent block. A blockchain is any block, and all of its parents recursively.
```

docs/conceptual/core/consensus.md
```diff
@@ -2,14 +2,104 @@
...
+
+## Finality
+Transactors in any system want to know when their transaction is finalized, and blockchain is no different. In some traditional systems, finality happens when a receipt is handed over, or papers are signed.
+
+Using the block authoring schemes and fork choice rules described so far, transactions are never truly finalized with certainty. There is always a chance that a longer (or heavier) chain will come along and revert your transaction. However, the more blocks are built on top of a particular block, the less likely it is to ever be reverted. In this way block authoring along with a proper fork choice rule provide probabilistic finality.
+
+When true finality is desired, an additional game, known as a finality gadget can be added to the blockchain's logic. In such a game, members of a fixed authority set cast finality votes, and when a certain threshold is reached, the block is deemed finalized, and will never be reverted.
```
4meta5: is that threshold >2/3n + 1 for n validators? should it be mentioned -- when I read "certain" threshold, I immediately wonder what the required threshold for finality is

JoshOrndorff: I had the same thought. the reason I didn't write it in yet, is because I wasn't sure whether there were other finality gadgets (possibly with weaker guarantees) that have a lower threshold. 2n/3 + 1 is definitely common and afaik ideal, but I can imagine inventing a finality gadget that requires 100% voting threshold.

Maybe it's worth mentioning that 2/3 is the most common and linking a BFT paper in the learn more section?

4meta5: Yes, would be nice to link the required lower bound for BFT finality.
```suggestion
When true finality is desired, an additional game, known as a finality gadget can be added to the blockchain's logic. In such a game, members of a fixed authority set cast finality votes for specific blocks. The threshold for achieving [Byzantine Fault Tolerance](https://people.eecs.berkeley.edu/~luca/cs174/byzantine.pdf) requires that greater than 2/3rds of the validator set's voting power commits to each finalized block. This quorum requirement for finality could also be configured to require more support for greater guarantees against reversion of finalized blocks.
```

4meta5: Also, I don't want to be *that* pedantic, but it's not really "true finality". @joepetrowski gave a great presentation on this a while back, I wonder what he thinks of this part

joepetrowski: There are a few problems here. I'm wary of adjectives and adverbs in general, especially ones like "true" and "always." There are indeed other protocols that require higher thresholds like 4/5. There is a proof that you can never do _better_ than 2/3, though. A higher threshold does not in any way provide a better guarantee (and therein you see the problem with the adjective "_true_ finality," you can't have something be "true" if its guarantees are on a sliding scale (and are they even guarantees then?)).

You should say "deterministic finality" and emphasize that it provides safety. You cannot re-org a finalized block without a hard fork - which requires some external coordination. The chain cannot simply re-org below the finalized head the way a chain with probabilistic finality can.

shawntabrizi:
@JoshOrndorff I did not look so deeply, but I generally agree that this is a good start to merge into the dev hub for now without over committing to large amounts of content around Grandpa/Babe etc...

But what I would like is a small "product" list of what is available in Substrate at the time of writing.

Something like the "virtual machines" list here:
https://substrate.dev/docs/en/next/conceptual/runtime/contracts/

Would just be 2-5 sentences on what the item is, what it generally offers, and why you would want to pick it or something like that.

docs/conceptual/core/consensus.md
```diff
@@ -2,14 +2,104 @@
...
-* What is consensus for on a blockchain?
+## State Machines and Conflicts
+A blockchain runtime is a [state machine](https://en.wikipedia.org/wiki/Finite-state_machine). It has some internal state, and state transition function that allows it to transition from its current state to a future state. In most runtimes there are states that have valid transitions to multiple future states, but a single transition must be selected. Consider the following examples.
 
-* Block Production
+### Voting Conflict
+A blockchain runtime tallys simple two-option votes for a  DAO. When a new election is opened, Alice, a member of the DAO, can vote Aye. Alice could also vote Nay. But Alice cannot vote both Aye and Nay in the same election.
 
-* Finality
+### Land Registry Conflict
```
shawntabrizi: Can this conflict be explained without "Land Registry"? Something more generic?

docs/conceptual/core/consensus.md
```diff
@@ -2,14 +2,104 @@
 title: Consensus
 ---
```
shawntabrizi: Please add a small 1-2 sentences telling the user what they will find on this page.

docs/conceptual/core/consensus.md
```diff
@@ -2,14 +2,104 @@
...
-* Block Production
+### Voting Conflict
+A blockchain runtime tallys simple two-option votes for a  DAO. When a new election is opened, Alice, a member of the DAO, can vote Aye. Alice could also vote Nay. But Alice cannot vote both Aye and Nay in the same election.
 
-* Finality
+### Land Registry Conflict
+Consider a blockchain runtime that serves as a land registry for a futuristic nation-state. The registry keeps track of transfers of land ownership among citizens, and occasionally opens new plots of land to be claimed. For a particular newly opened land run, Alice could claim the plot, or Bob could claim it. But they cannot both claim it.
 
-* NPoS
+### Classic Double Spend Conflict
```
shawntabrizi:
```suggestion
### Double Spend Conflict
```

shawntabrizi:
Please use a tool like https://marketplace.visualstudio.com/items?itemName=stkb.rewrap

It makes it easier to provide feedback on GitHub

docs/conceptual/core/consensus.md
```diff
@@ -2,14 +2,104 @@
 title: Consensus
 ---
 
-* What is consensus for on a blockchain?
+## State Machines and Conflicts
+A blockchain runtime is a [state machine](https://en.wikipedia.org/wiki/Finite-state_machine). It has some internal state, and state transition function that allows it to transition from its current state to a future state. In most runtimes there are states that have valid transitions to multiple future states, but a single transition must be selected. Consider the following examples.
 
-* Block Production
+### Voting Conflict
+A blockchain runtime tallys simple two-option votes for a  DAO. When a new election is opened, Alice, a member of the DAO, can vote Aye. Alice could also vote Nay. But Alice cannot vote both Aye and Nay in the same election.
```
kaichaosun:
```suggestion
A blockchain runtime tallys simple two-option votes for a DAO. When a new election is opened, Alice, a member of the DAO, can vote Aye. Alice could also vote Nay. But Alice cannot vote both Aye and Nay in the same election.
```

docs/conceptual/core/consensus.md
```diff
@@ -2,14 +2,104 @@
...
+### Classic Double Spend Conflict
+A pure cryptocurrency, like Bitcoin, tracks the minting and spending of tokens. Alice holds a particular coin, so she can pay it to Bob. She could also pay that coin to Carol. But she cannot pay the same coin to both Bob and Carol.
 
-* PoW
+The unifying property in these examples is that there are multiple transactions, each of which is individually valid, but are mutually exclusive. Among the alternatives, a single transaction must be chosen.
 
-* What would it take to write my own consensus?
+
+## Conflict Exclusion
+In centralized systems, the central authority chooses among mutually exclusive alternatives by recording transactions in the order it sees them, and choosing the first of the competing alternatives when a conflict arises. In decentralized systems, the nodes will see transactions in different orders, and thus they must use a more elaborate method to exclude transactions. As a further complication, blockchain networks strive to be fault tolerant, which means that they should continue to provide consistent data even if up to one third of participants are not following the rules.
```
kaichaosun: `one third` rule only applied to BFT, but it may not a good idea to detail such principle, instead we can explicitly mention Substrate default consensus (GRANDPA) here.

docs/conceptual/core/consensus.md
```diff
@@ -2,14 +2,104 @@
...
+
+One practical solution, and the one taken in Substrate, is to find a decentralized way to agree on transaction ordering, and then resolve conflicts by selecting the first alternative seen just as the centralized authority did.
+
+> In general a blockchain just needs some method to choose among conflicting transactions, and there are interesting paradigms that do not involve a total ordering. Such methods will not be considered further here.
+
+## Fork Choice Rules
+
+A block contains many transactions (technically it contains zero or more transactions), and a reference to its parent block. A blockchain is any block, and all of its parents recursively.
+
+The first step to resolving conflicting transaction is insisting that no blockchain can ever have two conflicting transactions. If a block comes along that conflicts with one of its ancestors, that block is invalid, and the nodes in the network will reject it.
```
kaichaosun:
```suggestion
The first step to resolve the conflicting transactions is insisting that no blockchain can ever have two conflicting transactions. If a block comes along that conflicts with one of its ancestors, that block is invalid, and the nodes in the network will reject it.
```

docs/conceptual/core/consensus.md
```diff
@@ -2,14 +2,104 @@
...
+
+Nodes that are following the rules of the consensus protocol always produce new blocks on top of the chain that the fork choice rule tells them to. But blockchains still must tolerate nodes that cheat and build blocks in the wrong place. Considering the example fork choice rules we studied above, it's easy to see that an attacker who wanted to revert many blocks worth of transactions could do so by building a longer chain.
+
+To prevent malicious nodes from creating attack chains and overwhelming the honest chain whenever they please, block production must be throttled so that nodes can only produce blocks at a given rate. There are two common ways of achieving this throttling.
+
+### Proof of Work
+In proof of work systems like Bitcoin, any node may produce a block at any time, so long as they have solved a computationally-intensive problem. Solving the problem takes cpu time, and thus miners can only produce blocks in proportion with their computing resources. Substrate provides a Proof of Work consensus engine.
```
kaichaosun:
```suggestion
In proof of work systems like Bitcoin, any node may produce a block at any time, so long as it has solved a computationally-intensive problem. Solving the problem takes CPU time, and thus miners can only produce blocks in proportion with their computing resources. Substrate provides a Proof of Work consensus engine.
```

docs/conceptual/core/consensus.md
```diff
@@ -2,14 +2,104 @@
...
+
+### Slots
+When using a slot based consensus algorithm, there must be a known set of validators who are permitted to produce blocks. Time is divided up into discrete slots, and during each slot only some of the validators may produce a block. The specifics of which validators can author blocks during each slot vary from engine to engine. Substrate provides Aura and Babe, both of which are slot-based block authoring engines.
+
+## Finality
+Transactors in any system want to know when their transaction is finalized, and blockchain is no different. In some traditional systems, finality happens when a receipt is handed over, or papers are signed.
```
kaichaosun: `Transactors` looks not familiar to me, how about `users`?

docs/conceptual/core/consensus.md
kaichaosun: @4meta5 I appreciate it if you can share the link of the presentation.

docs/conceptual/core/consensus.md
```diff
@@ -2,14 +2,104 @@
...
+
+When true finality is desired, an additional game, known as a finality gadget can be added to the blockchain's logic. In such a game, members of a fixed authority set cast finality votes, and when a certain threshold is reached, the block is deemed finalized, and will never be reverted.
+
+> There are systems that couple block authoring more tightly with finality by considering the act of building a block as a finality vote for that block. However the two can always be separated.
+
+In systems that use a finality gadget, the fork choice rule must be modified to consider the results of the finality game. For example, instead of taking the longest chain period, a node would take the longest chain that contains the most recently finalized block. Substrate provides the Grandpa finality gadget.
```
kaichaosun:
```suggestion
In systems that use a finality gadget, the fork choice rule must be modified to consider the results of the finality game. For example, instead of taking the longest chain period, a node would take the longest chain that contains the most recently finalized block. Substrate provides the GRANDPA finality gadget.
```

docs/conceptual/core/consensus.md
JoshOrndorff: Land registry is an example. IT sits alongside the voting example and the double spend example.

docs/conceptual/core/consensus.md
JoshOrndorff: I'm open to that suggestion. Do others agree either way? I had originally chosen the term transactors to be as specific as possible (rule out those who use the blockchain for forensics, or only as full nodes, etc).

JoshOrndorff: Actually, I just took the suggestion. If anyone agrees with my reason for using "transactors" we can switch back. Until then, let's keep this moving.

docs/conceptual/core/consensus.md
JoshOrndorff: I've removed _this_ occurrence of the specific threshold, saving that conversation for later. When discussing GRANDPA I mentioned it again, and in the learn more section I link some resources that explain it.

PR #343, opened 2019-11-22
Fix class warning

shawntabrizi:
Fixes a small warning error about `class` vs `className`


shawntabrizi:
I searched `class` and looked in `jsx` files

PR #345, opened 2019-11-25
Tutorials frame pallet update

riusricardo:
Updates: 

- SRML to FRAME
- srml- to pallet-
- Add a module tutorial

File names and links will be updated in another PR.


docs/tutorials/creating-a-runtime-module.md
```diff
@@ -1,20 +1,20 @@
 ---
-title: "Creating a Runtime Module"
+title: "Creating a Runtime Pallet"
```
shawntabrizi: AFAIK it should be `Runtime Module` -> `Pallet` rather than `Runtime Pallet`.

Can you confirm with @gavofyork in some public place?

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -96,20 +96,20 @@ To see how these features actually get used in the runtime code, we can open the
 include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 
 use rstd::prelude::*;
-//--snip--
```
JoshOrndorff: Definitely not a blocker, but why prefer this format?

I think we chose `//--snip--` following in the footsteps of the rust book eg https://doc.rust-lang.org/book/ch19-04-advanced-types.html#the-never-type-that-never-returns

riusricardo: You are right. I just did it because it was clear to do it this way after comments. Like in this example:
```rust
// These time units are defined in number of blocks.
   /* --snip-- */
```

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -172,83 +172,111 @@ Now that we have successfully imported the Contracts module crate, we need to ad
...
 
 ### Adding Contract to the Construct Runtime Macro
 
-Next, we need to add the module to the `construct_runtime!` macro. For this, we need to determine the types that the module exposes so that we can tell the our runtime that they exist. The complete list of possible types can be found in the [`construct_runtime!` macro documentation](https://substrate.dev/rustdocs/master/palette_support/macro.construct_runtime.html).
+Next, we need to add the pallet to the `construct_runtime!` macro. For this, we need to determine the types that the pallet exposes so that we can tell the our runtime that they exist. The complete list of possible types can be found in the [`construct_runtime!` macro documentation](https://substrate.dev/rustdocs/master/palette_support/macro.construct_runtime.html).
 
-If we look at the Contracts module in detail, we know it has:
+If we look at the Contracts pallet in detail, we know it has:
 
 * Module **Storage**: Because it uses the `decl_storage!` macro.
```
satyamakgec: I think here `Module` into replacing with pallet

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -172,83 +172,111 @@ Now that we have successfully imported the Contracts module crate, we need to ad
...
 * Module **Event**s: Because it uses the `decl_event!` macro.
 * **Call**able Functions: Because it has dispatchable functions in the `decl_module!` macro.
 * **Config**uration Values: Because the `decl_storage!` macro has `config()` parameters.
-* The **Module** type from the `decl_module!` macro.
+* The **pallet** type from the `decl_pallet!` macro.
```
satyamakgec: Right now every pallet has `decl_module!` macro instead of the `decl_pallet!` macro so I am assuming this will be future changes as the master branch doesn't reflect `decl_pallet!` right now but it is a schedule change.

satyamakgec: Or it is a mistake to have `decl_pallet!`, `It should be replaced with decl_module!`

riusricardo: Nice! you are right, thanks for the catch.

docs/tutorials/creating-a-runtime-module.md
```diff
@@ -108,51 +108,53 @@ std = [
 
 ### Consistent Substrate Dependencies
 
-All Substrate modules will depend on some low-level module libraries such as
-`srml-system` and `srml-support`. These libraries are pulled from the main
+All Substrate pallets will depend on some low-level pallet libraries such as
```
shawntabrizi:
```suggestion
All Substrate pallets will depend on some low-level FRAME libraries such as
```

docs/tutorials/creating-your-first-substrate-chain/build.md
```diff
@@ -156,7 +156,7 @@ decl_module! {
             // Verify that the specified proof has not been claimed yet or error with the message
             ensure!(!Proofs::<T>::exists(&proof), "This proof has already been claimed.");
 
-            // Call the `system` runtime module to get the current block number
+            // Call the `system` runtime pallet to get the current block number
```
shawntabrizi:
```suggestion
            // Call the `system` pallet to get the current block number
```

docs/tutorials/creating-a-runtime-module.md
shawntabrizi: What did he say?

PR #352, opened 2019-11-28
Remove crates.parity.io links

shawntabrizi:
Most of them convert to `rustdocs/v1.0`, one case, I changed to `rustdocs/master/pallet_`


JoshOrndorff:
Just so we all understand the vision, for which of the following reasons are we hosting the rustdocs under substrate.dev as well as crates.parity.io?

* So substrate.dev is more self-contained
* So we can host docs that correspond to released version (like v1.0) while crates.parity.io follows master
* Other

Is the intention always to host a set of docs that follows master? Or are we only doing that until v1.0 is released?

PR #357, opened 2019-12-03
Update add a module tutorial

riusricardo:


JoshOrndorff:
Looking good so far. It might be worth mentioning that adding a runtime API is not needed for most modules, and is technically not _needed_ for the contracts module either. But it gains us the features described here https://github.com/paritytech/substrate/pull/3563

riusricardo:
Good point @JoshOrndorff, it's worth mentioning that. I'm just missing the explanations but the code is up to date with Substrate master. This pallet became more complex to integrate than many others.

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -317,16 +335,106 @@ To enable this, we simply need to add `Contracts` type that we defined in the `c
...
+                use contracts_rpc_runtime_api::{GetStorageError as RpcGetStorageError};
+                /// Map the contract error into the RPC layer error.
+                match rpc_err {
+                    GetStorageError::ContractDoesntExist => RpcGetStorageError::ContractDoesntExist,
+                    GetStorageError::IsTombstone => RpcGetStorageError::IsTombstone,
+                }
+            })
+        }
+    }
+   /*** ***/
```
JoshOrndorff: What is this empty comment for?

riusricardo: I'm trying to use it as a way to indicate the end of a block. It has the same pattern as the `/*** Add this block ***/`. If it is not clear I can change/remove it.

JoshOrndorff: I see what you mean now. Maybe consider `/*** End Added Block ***/`?? 

I'll leave it up to you.

riusricardo:
I need to do another review and probably fixes introduced by `td-everyone-will-hate-me-for-this` branch.

JoshOrndorff:
Here's a PR for node template that incorporates those changes https://github.com/substrate-developer-hub/substrate-node-template/pull/14. If you check this tutorial against the new node template, could you review that PR too?

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -448,20 +448,20 @@ Then inside the `testnet_genesis` function we need to add the contract configura
 fn testnet_genesis(initial_authorities: Vec<(AuraId, GrandpaId)>,
     root_key: AccountId,
     endowed_accounts: Vec<AccountId>,
-    enable_println: bool) -> GenesisConfig {
-    /*** Add this block ***/
+    _enable_println: bool) -> GenesisConfig {
```
JoshOrndorff: Why is this variable name preceded with an underscore?

I believe that's recommended for unused variables, but this one is used a few lines below.

riusricardo: It is marked as unused by default. So I just took the definition that comes on the template. I can also point this out and make it part of the required changes.

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -317,44 +335,190 @@ To enable this, we simply need to add `Contracts` type that we defined in the `c
...
 
-The last thing we need to do in order to get your node up and running is to establish a genesis configuration for the Contracts pallet. Not all pallets will have a genesis configuration, but if they do, you can use its documentation to learn about it. For example, [`pallet_contracts::GenesisConfig` documentation](https://substrate.dev/rustdocs/master/pallet_contracts/struct.GenesisConfig.html) describes all the fields you need to define for the Contracts pallet. This definition is controlled in `substrate-node-template/src/chain_spec.rs`. We need to modify this file to include the `ContractsConfig` type and add the contract price units at the top:
+Not all pallets will have a genesis configuration, but if they do, you can use its documentation to learn about it. For example, [`pallet_contracts::GenesisConfig` documentation](https://substrate.dev/rustdocs/master/pallet_contracts/struct.GenesisConfig.html) describes all the fields you need to define for the Contracts pallet. This definition is controlled in `substrate-node-template/src/chain_spec.rs`. We need to modify this file to include the `ContractsConfig` type and the contract price units at the top:
```
JoshOrndorff: This bit about the GenesisConfig falls right in the middle of installing the RPC. What if you switch the order to.

* Exposing the contracts RPC API
* Install the RPC in the node's service (This is like the second half of exposing the RPC)
* Genesis Configuration

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -317,44 +335,190 @@ To enable this, we simply need to add `Contracts` type that we defined in the `c
...
 **`src/chain_spec.rs`**
 
 ```rust
-use runtime::{ContractsConfig, MILLICENTS};
+use node_template_runtime::{ContractsConfig, MILLICENTS};
 ```
 
 Then inside the `testnet_genesis` function we need to add the contract configuration to the returned `GenesisConfig` object as followed:
 
+> Note: We are taking the value `_enable_println` from the function parameters.
```
JoshOrndorff: _variable_name

JoshOrndorff: Oh, I see you commented to remove the underscore :+1:

docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -317,44 +335,190 @@ To enable this, we simply need to add `Contracts` type that we defined in the `c
...
+    type OnFreeBalanceZero = Contracts;
     /* --snip-- */
 }
 ```
-
 Now, when the Balances pallet detects that the free balance of an account has reached zero, it calls the `on_free_balance_zero` function of the Contracts pallet.
 
+### Exposing The Contracts API
+
+We now need to enable calls to the contracts pallet. Not all of the pallets require using a specialized configuration to interact with them. However, the contracts pallet requires these API to call the sandboxed contracts.
```
JoshOrndorff: I don't know whether this is true, but myunderstanding, was that this RPC was just for doing dry-run contract calls and reading contract storage from offchain, and it is still possible to make actual on-chain calls without the RPC. If that is indeed the case, I think this sentence is misleading.

riusricardo: Fair point it is part true but it requires more clarification. As you said it works to call the local node storage instead of making a transaction to get the data.

riusricardo:
I'm gonna test again because it could be that polkadot.js.org/apps has an issue with the version that we are using for the templates.

dtyoo:
@JoshOrndorff thanks for pointing me over here and @riusricardo for putting it together.  I hit a couple issues going through it.  Still learning so not sure if this is something specific to my setup or something else, but thought I would comment here with my findings.  I hit the following errors when using these instructions against the recently updated substrate-node-template (rev: 40a16efefc070faf5a25442bc3ae1d0ea2478eee):

```
error[E0432]: unresolved import `contracts_rpc_runtime_api`
  --> /home/derek/moonbeam/runtime/src/lib.rs:14:5
   |
14 | use contracts_rpc_runtime_api::ContractExecResult;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared type or module `contracts_rpc_runtime_api`
...snip...
```

To solve I removed the pallet prefix in runtime/Cargo.toml resulting in
```
'contracts-rpc-runtime-api/std',
```
and
```
[dependencies.contracts-rpc-runtime-api]
```

Next I hit this:

```
error[E0425]: cannot find value `enable_println` in this scope
   --> src/chain_spec.rs:130:53
    |
130 |     contracts_config.current_schedule.enable_println = enable_println;
    |                                                        ^^^^^^^^^^^^^^ help: a local variable with a similar name exists: `_enable_println`
```
To solve I added an underscore to the enable_println variable in chain_spec.rs resulting in:
```
contracts_config.current_schedule.enable_println = _enable_println;

```

After successfully building I can deploy and call flipper functions from the current polkadot-js apps!

riusricardo:
@dtyoo Thanks for the catch. I thought that I changed the `pallet-contracts..` definition. It looks like I forgot it here. It was a super recent change that mixed some references.
I'll push a fix. 

About the `_enable_println` there is a `Note` that tells you to remove the underscore in the function parameter.
>Note: We are taking the value _enable_println from the function parameters. Make sure to remove the underscore that precedes the parameter definition.

I'll make it more clear :)

PR #378, opened 2019-12-18
Fix pallet-contracts

riusricardo:
Fixes `contracts-rpc-runtime-api` import


riusricardo:
yeah, the problem is that there are not named the same in `node`and in `node-template` after the last rename. Until that gets figured out, we should not mess that much.

PR #379, opened 2019-12-19
Off-chain workers docs

jimmychu0807:


jimmychu0807:
@joepetrowski  @JoshOrndorff  @riusricardo 

I am done with my doc on off-chain worker. `Conceptual` part is mostly from v1 as it is still relevant. There are still some stuff in `development` part I am missing I need to further explore. They are commented at the end (they are not visible in the website) But see if the current content is good enough to publish and do some good. 

Please make improvement on my grammar.

Thanks.

docs/conceptual/core/off-chain-workers.md
```diff
@@ -0,0 +1,30 @@
...
+
+To make the off-chain data integration secure and more efficient, Substrate provides off-chain workers. The off-chain worker subsystem allows execution of long-running and possibly non-deterministic tasks (e.g. web requests, encryption/decryption and signing of data, random number generation, CPU-intensive computations, enumeration/aggregation of on-chain data, etc.) which could otherwise require longer than the block execution time.
+
+Off-chain workers have their own Wasm execution environment outside of the Substrate runtime. This separation of concerns is to make sure that the block production is not impacted by the long-running tasks. However, as the off-chain workers are declared in the same code as the runtime, they can easily access on-chain state for their computations.
+
+![Off-chain Workers](/docs/assets/off-chain-workers-02.png)
+
+## APIs
+
+Additionally, they have these capabilities to communicate with the external world via extended set of APIs.
```
JoshOrndorff:
```suggestion
Offchain workers have access to extended APIs for communicating with the external world:
```

docs/conceptual/core/off-chain-workers.md
```diff
@@ -0,0 +1,30 @@
...
+
+* Ability to submit transactions (either signed or unsigned) back on-chain to publish computation results.
+* A fully-featured HTTP client allowing to access and fetch data from external services.
+* Access to the local keystore to sign and verify statements or transactions.
+* An additional, local key-value database shared between all off-chain workers.
+* A secure, local entropy source for random number generation.
```
JoshOrndorff: I almost removed the comma, but I think it's actually correct as it stands https://www.quickanddirtytips.com/education/grammar/commas-with-adjectives

docs/conceptual/core/off-chain-workers.md
```diff
@@ -0,0 +1,30 @@
...
+* Access to local node precise time and ability to sleep/resume work.
```
JoshOrndorff:
```suggestion
* Access to the node's precise local time and the ability to sleep and resume work.
```

docs/conceptual/core/off-chain-workers.md
```diff
@@ -0,0 +1,30 @@
...
+
+Off-chain workers can be initiated from within a special function in your runtime implementation, `fn offchain_worker(block: T::BlockNumber)`. The function is executed after each block import. To communicate results back to the chain, off-chain workers can submit signed or unsigned transactions that are later to be included in subsequent blocks.
```
JoshOrndorff:
```suggestion
Off-chain workers can be initiated from within a special function in your runtime implementation, `fn offchain_worker(block: T::BlockNumber)`. The function is executed after each block import. To communicate results back to the chain, off-chain workers can submit signed or unsigned transactions to be included in subsequent blocks.
```

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
+---
+title: Off-Chain Workers
+---
+
+You can learn about the high level concept of off-chain workers in our [Conceptual Guide](conceptual/core/off-chain-workers.md).
```
JoshOrndorff:
```suggestion
This article covers the technical aspects of using off-chain workers in a Substrate runtime. For a conceptual overview of off-chain workers see [Conceptual Guide](conceptual/core/off-chain-workers.md).
```

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+```rust
+//For better debugging (printout) support
+use support::{ debug, dispatch };
+use system::offchain;
+use sp_runtime::transaction_validity::{
+  TransactionValidity, TransactionLongevity, ValidTransaction, InvalidTransaction
+};
+```
+
+Include the following associated type in your runtime trait for sending signed and unsigned transactions from off-chain worker.
```
JoshOrndorff:
```suggestion
Include the following associated types in your pallet's configuration trait for sending signed and unsigned transactions from an off-chain worker.
```

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+title: Off-Chain Workers
+---
+
+You can learn about the high level concept of off-chain workers in our [Conceptual Guide](conceptual/core/off-chain-workers.md).
+
+## Using Off-Chain Workers in Runtime
+
+To use off-chain workers in your runtime, include these following modules
+
+**runtime/src/my_runtime.rs**
```
JoshOrndorff: For tutorials I like the idea of providing a specific path to the file being modified. I'm not sure it makes sense here since we don't know the name of the reader's pallet or the filesystem location where it is being developed. It may be in its own rust module rather than right in the runtime.

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+  /// The overarching event type.
+  type Event: From<Event<Self>> + Into<<Self as system::Trait>::Event>;
+  type Call: From<Call<Self>>;
+
+  type SubmitSignedTransaction: offchain::SubmitSignedTransaction<Self, <Self as Trait>::Call>;
+  type SubmitUnsignedTransaction: offchain::SubmitUnsignedTransaction<Self, <Self as Trait>::Call>;
+}
+```
+
+Now inside the `decl_module!` dispatch functions section, define the following function and this serves as the entry point of the off-chain worker after every block import.
```
JoshOrndorff:
```suggestion
Now inside the `decl_module!` section, define the `offchain_worker` function. This function serves as the entry point of the off-chain worker and runs after every block import.
```

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+    // ...
+    // ...
+    fn offchain_worker(block: T::BlockNumber) {
+      debug::info!("Hello World.");
+    }
+  }
+}
+```
+
+By default, the offchain worker doesn't have direct access to user keys (even in development environment), but can only access  app-specific subkeys for security reason. You need to define the `KeyTypeId` at the top of your runtime that is used to group your app-specific subkeys as followed.
```
JoshOrndorff:
```suggestion
By default, the offchain worker doesn't have direct access to user keys (even in development environment), but can only access app-specific subkeys for security reason. You need to define the `KeyTypeId` at the top of your runtime that is used to group your app-specific subkeys as followed.
```

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+// ...
+// ...
+pub mod crypto {
+  pub use super::KEY_TYPE;
+  use sp_runtime::app_crypto::{app_crypto, sr25519};
+  app_crypto!(sr25519, KEY_TYPE);
+}
+```
+
+To support this runtime, implement in your library module
```
JoshOrndorff:
```suggestion
As with any other pallet, your runtime must implement the pallet's configuration trait.
```

This is starting to feel more like a tutorial or recipe.

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+
+**runtime/src/lib.rs**
+
+```rust
+/// Define the Transaction signer using the Key definition
+type SubmitTransaction = system::offchain::TransactionSubmitter<
+  my_runtime::crypto::Public, Runtime, UncheckedExtrinsic>;
+
+impl runtime::Trait for Runtime {
```
JoshOrndorff: The pallet should not be called `runtime`. How about something like `my_pallet` or `offchain_pallet`?

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+
+    // For just using signed transactions, it can just well be:
+    // MyRuntime: my_runtime::{ Module, Call, Storage, Event<T> }
+  }
+);
+```
+
+## Insert Keys in `service.rs`
+
+We have specify a local keystore with `KeyTypeId` to store app-specific sub-keys that are accessible by off-chain workers for signing transactions. But currently there is no sub-keys inside. We can add it in one of the following two ways.
```
JoshOrndorff:
```suggestion
We have specified a local keystore with `KeyTypeId` to store app-specific keys that are accessible by the off-chain worker for signing transactions. But currently there are no keys inside. We can add keys in one of the following two ways.
```

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+In development environment, you can add the first user key as the app sub-key. Update the `service.rs` as followed.
+
+**node/src/service.rs**
+
+```rust
+pub fn new_full<C: Send + Default + 'static>(config: Configuration<C, GenesisConfig>)
+  -> Result<impl AbstractService, ServiceError>
+{
+  // ...
+  // ...
```
JoshOrndorff: // --snip--

I believe we've been using this `// --snip--` syntax to follow the rust book.

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+    service
+      .keystore()
+      .write()
+      .insert_ephemeral_from_seed_by_type::<runtime::my_runtime::crypto::Pair>(
+        &seed,
+        runtime::my_runtime::KEY_TYPE,
+      )
+      .expect("Dev Seed should always succeed.");
+  }
+}
```
JoshOrndorff: What key or keys actually get added here? Just Alice? Alice, Bob, etc? Something else?

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+```
+
+Then you would be able to sign transactions, and this is good for **DEVELOPMENT only**.
+
+### Add an App Subkey via CLI
+
+In a more realistic setting, after setting up your substrate node, you can add a new app subkey via
```
JoshOrndorff:
```suggestion
In a more realistic setting, after setting up your Substrate node, you can add a new app subkey via
```

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+the command line interface as followed.
```
JoshOrndorff:
```suggestion
the command line interface as follows.
```

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+
+```bash
+# Generate a new account
+$ subkey -s generate
+
+# Submit a new key via RPC
+$ http localhost:9933 jsonrpc=2.0 id=1 method=author_insertKey params:='["<YourKeyTypeId>", "<YourSeedPhrase>", "YourPublicKey"]'
```
JoshOrndorff: My debian system does not come with `http` installed. Maybe `curl` is more popular? Or this can be done through Apps UI. I wonder if we should have a separate doc on making RPC calls?

This comment is not a blocker. I'm just thinking out loud and I think competent developers will be able to take what you've written and figure out how to make the RPC call.

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+    }
+  }
+}
+```
+
+Here, after having defined the on-chain callback function, in the off-chain worker you can specify
+that function to be called back in the next block import phase. You then submit a signed transaction
+to the node.
+
+If you look at the implementation of `fn system::offchain::submit_signed ()` in [Substrate codebase](https://github.com/paritytech/substrate), you would realized it is calling the on-chain callback for each of the key in local keystore.
```
JoshOrndorff: This link should go to a specific place, not just the main Substrate repo.
```suggestion
If you look at the implementation of `fn system::offchain::submit_signed ()` in the [Substrate codebase](https://github.com/paritytech/substrate), you will see it is calling the on-chain callback for each of the key in local keystore.
```

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+      // Here we specify the function to be callback on-chain in next block import
+      let call = Call::onchain_callback(block, b"hello world!".to_vec());
+      T::SubmitUnsignedTransaction::submit_unsigned(call);
+    }
+  }
+}
+```
+
+By default, all unsigned transactions are treated as invalid transactions. You need to add the following
+code piece in `my_runtime.rs` to whitelist some on-chain functions that are allowed to be called with
```
JoshOrndorff:
```suggestion
code piece in `my_runtime.rs` to explicitly allow some on-chain functions to be called with
```

Stealing this suggestion from Fuchsia docs https://fuchsia.dev/fuchsia-src/contribute/best-practices/respectful_code#what_are_examples_of_terminology_to_be_avoided

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,383 @@
...
+}
+```
+
+You see that we add a `deprecated` attribute to prevent warning message from display. This is because
+this part of the API is still in transition and will be updated in coming Substrate release.
+
+## Parameters in On-Chain Callbacks
+
+When making a on-chain callback, our implementation hashes the function name together with all parameter
+values and store it to be called in next import. If we find that the hash value existed, meaning a function with the same parameter value set has been called before, for signed transaction this causes a function replacement if this call is set with a higher priority; and for unsigned transaction, this callback is simply ignored.
```
JoshOrndorff:
```suggestion
values and stores it to be called in next import. If we find that the hash value existed, meaning a function with the same parameter value set has been called before, for signed transaction this causes a function replacement if this call is set with a higher priority; and for unsigned transaction, this callback is simply ignored.
```

jimmychu0807:
@JoshOrndorff 
> I wonder if it would be better as a recipe or tutorial? @jimmychu0807 I would love to have your price-fetching example as part of the recipes.

I am updating the guide now. I will later add an additional chapter in  recipe with the price-fetching example.

docs/development/module/off-chain-workers.md
jimmychu0807: True. But the filename provide the context where the code to be added. I was thinking,

1. skipping the filename altogether - then readers may not know where to add the code. Since some code are added in `lib.rs`, and some in the `runtime.rs`

2. `runtime.rs` (or `pallet.rs`) - this is the best name I can thinking of without the file path

3. `runtime/src/my_runtime.rs` - giving the path, hopefully it gives more clue on where the code to be added

What do you think?

docs/development/module/off-chain-workers.md
jimmychu0807: Referring to: https://github.com/paritytech/substrate/blob/master/client/cli/src/lib.rs#L901-L908

In development, it is only Alice.

docs/development/module/off-chain-workers.md
jimmychu0807: Doing a simple doc/guide on RPC call will be great.

Just realized using `curl` is quite cumbersome, see the update. But this should be more widely-runnable to readers.

docs/development/module/off-chain-workers.md
jimmychu0807: Using `offchain_pallet` now.

jimmychu0807:
@JoshOrndorff I have incorporated all comments except the one about removing the full path name of files. I am not sure what the best solution is. The current path name give readers most context where to add the code.

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,416 @@
...
+
+You see that we add a `deprecated` attribute to prevent warning message from display. This is because
+this part of the API is still in transition and will be updated in coming Substrate release.
+
+## Parameters in On-Chain Callbacks
+
+When making a on-chain callback, our implementation hashes the function name together with all
+parameter values and stores it to be called in next import. If we find that the hash value exists,
+meaning a function with the same parameter value set has been called before, for signed transaction
+this causes a function replacement if called with a higher priority; for unsigned transaction, this
```
riusricardo:
```suggestion
, the function call will be replaced if it is called with a higher priority; for an unsigned transaction, this
```

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,416 @@
...
+callback is simply ignored.
+
+If your pallet is making on-chain callbacks regularly and you expect it will have duplicated
```
riusricardo:
```suggestion
If your pallet is making on-chain callbacks regularly and you expect it to have duplicate
```

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,416 @@
...
+parameter value set every once in a while, you can always pass in an additional parameter of the
```
riusricardo:
```suggestion
sets of parameters every once in a while, you can always pass in an additional parameter of the
```

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,416 @@
...
+        longevity: TransactionLongevity::max_value(),
+        propagate: true,
+      }),
+      _ => InvalidTransaction::Call.into()
+    }
+  }
+}
+```
+
+You see that we add a `deprecated` attribute to prevent warning message from display. This is because
```
riusricardo:
```suggestion
You see that we add a `deprecated` attribute to prevent warning messages from being displayed. This is because
```

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,416 @@
...
+
+When making a on-chain callback, our implementation hashes the function name together with all
+parameter values and stores it to be called in next import. If we find that the hash value exists,
+meaning a function with the same parameter value set has been called before, for signed transaction
+this causes a function replacement if called with a higher priority; for unsigned transaction, this
+callback is simply ignored.
+
+If your pallet is making on-chain callbacks regularly and you expect it will have duplicated
+parameter value set every once in a while, you can always pass in an additional parameter of the
+current block number, that is being passed in from `fn offchain_worker()`. This number is going to
```
riusricardo:
```suggestion
current block number thorough the `fn offchain_worker()` function. This number is going to
```

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,416 @@
...
+  }
+}
+```
+
+You see that we add a `deprecated` attribute to prevent warning message from display. This is because
+this part of the API is still in transition and will be updated in coming Substrate release.
+
+## Parameters in On-Chain Callbacks
+
+When making a on-chain callback, our implementation hashes the function name together with all
```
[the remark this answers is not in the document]
jimmychu0807: Agree with the `an`. For the second modification, I am more on the side of using present tense.

docs/development/module/off-chain-workers.md
JoshOrndorff: Yeah, I agree. Leaving some kind of path is better than no path.

I don't like the name `runtime.rs` for the pallet though because the complete runtime actually lives in `lib.rs`

jimmychu0807: True. Actually I have changed all file paths to `runtime/src/offchain_pallet.rs`.

Refer to this commit: 
https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/379/commits/90316e7925704c376a1012fdcb874a87280bfab2

See if the PR is good for approval. Thanks.

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,402 @@
...
+
+Now you are ready to to make a signed transaction from the off-chain worker. Go back to your pallet
+in `my_offchain_worker.rs`.
+
+```rust
+decl_module! {
+  pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+    // --snip--
+
+    // This is your regular on-chain extrinsics
```
joepetrowski: This doesn't make sense to me.

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,402 @@
...
+## Unsigned Transactions
+
+With the following code, you are able to send an unsigned transaction back to the chain.
+
+```rust
+decl_module! {
+  pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+    // --snip--
+
+    // This is your regular on-chain extrinsics
```
joepetrowski: Ditto

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,402 @@
...
+decl_module! {
+  // --snip--
+}
+
+impl<T: Trait> Module<T> {
+  // --snip--
+}
+
+#[allow(deprecated)]
+impl<T: Trait> support::unsigned::ValidateUnsigned for Module<T> {
```
joepetrowski: This should be removed ASAP, including the paragraph that follows this code snippet. Unfortunately, it seems we don't actually have a code example of using `SignedExtension` to do this. At a minimum, we should like people to the conceptual-extrinsics doc where we talk about validate-unsigned.

I think we can probably merge now with the example so that we don't leave people hanging, but should really caution them with "don't make this a habit."

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,402 @@
...
+}
+```
+
+You see we add a `deprecated` attribute to prevent warning messages from being displayed. This is
+because this part of the API is still in transition and will be updated in coming Substrate release.
+
+## Parameters in On-Chain Callbacks
+
+When making an on-chain callback, our implementation hashes the function name together with all
+of its parameter values. The callback will be stored and called during the next block next import.
```
riusricardo:
```suggestion
of its parameter values. The callback will be stored and called during the next block import.
```

docs/development/module/off-chain-workers.md
jimmychu0807: Updated, and added links at the bottom of each section, and the last `Next Steps` section.

jimmychu0807:
> This is really close, just a few changes to make first. A couple image updates:
> 

@joepetrowski  I end up unified using image with dotted red line. For v1 off-chain workers doc, I am keeping the original image as I am not sure if it was designed like that before. For v2, I have updated based on your feedback.

docs/development/module/off-chain-workers.md
```diff
@@ -0,0 +1,408 @@
...
+      }),
+      _ => InvalidTransaction::Call.into()
+    }
+  }
+}
+```
+
+We add a `deprecated` attribute to prevent warning messages from being displayed. It is
+because this part of the API is still in transition and will be updated in coming Substrate release.
+Please use this in caution for now.
```
joepetrowski:
```suggestion
Please use this with caution for now.
```

PR #380, opened 2019-12-19
New Hompage Structure

thisisvilla:
Following wireframe: https://paritytech.github.io/wireframe/devhub-home

@joepetrowski


website/static/js/old-pages/index.js
```diff
@@ -0,0 +1,264 @@
...
+          <li>
+            <translate>Hot-swappable consensus layer.</translate>
+          </li>
+          <li>
+            <translate>
+              Customizable transaction queue management system.
+            </translate>
+          </li>
+          <li>
+            <translate>Diverse library of runtime modules.</translate>
```
JoshOrndorff:
```suggestion
            <translate>Rich framework for composing runtime logic.</translate>
```

The term "runtime module" has been replaced with "pallet". I didn't suggest a direct translation because I feared people reading that far down the landing page wouldn't know what a pallet was yet.

joepetrowski: We may also want to avoid direct translations of pallet as it's also a nod to "crates" and that wordplay probably doesn't translate well.

thisisvilla: hey there! i noticed that change was suggested to the old home page. I went ahead and added it to the new homepage and also removed the blast banner as per the suggestion from @riusricardo. 

Also, just FYI, I haven't added the strings for translation just yet. That will be in in a future pull request. @ericmorel suggested it was cool to just publish the english version so the new page could be up sooner.

I'm making a pull request now for the changes listed above, and will make another pull request shortly for the new seminar page :) 

Let me know if you guys think of any other changes! Thanks!

PR #381, opened 2019-12-19
Fix order in tutorial

riusricardo:
Moves the `cargo run` instruction to the end.

### DIFF
--- a/docs/tutorials/adding-a-module-to-your-runtime.md
+++ b/docs/tutorials/adding-a-module-to-your-runtime.md
@@ -526,6 +526,8 @@ fn testnet_genesis(initial_authorities: Vec<(AuraId, GrandpaId)>,
 }
 ```
 
+## Start Your Upgraded Chain
+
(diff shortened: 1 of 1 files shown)


docs/tutorials/adding-a-module-to-your-runtime.md
```diff
@@ -533,22 +526,26 @@ fn testnet_genesis(initial_authorities: Vec<(AuraId, GrandpaId)>,
 }
 ```
 
-Note that you can tweak these numbers to your needs, but these are the values set at the [genesis configuration of the main Substrate node](https://github.com/paritytech/substrate/blob/master/bin/node/cli/src/chain_spec.rs).
+Now you are ready to compile and run your contract-capable node. We first need to purge the chain to remove the old runtime logic and have the genesis configuration initialized for the Contracts pallet. It is possible to upgrade the chain without purging it but it will remain out of scope for this tutorial.
```
JoshOrndorff: I agree this is a nicer place to have them run the result. I almost feel like this paragraph deserves it's own title so that the "finally, it all works" moment isn't buried under the heading `## Genesis Configuration`

```suggestion
## Start Your Upgraded Chain

Now you are ready to compile and run your contract-capable node. We first need to purge the chain to remove the old runtime logic and have the genesis configuration initialized for the Contracts pallet. It is possible to upgrade the chain without purging it but it will remain out of scope for this tutorial.
```

Also line length.

PR #390, opened 2019-12-28
Update First Chain Tutorial

riusricardo:
Updates tutorial to the same commit as the node-template.


riusricardo:
Fair point @kaichaosun. I was not aware of that repo. I'll do a PR to update it.

PR #403, opened 2020-01-06
Table of Contents: Remove Conceptual and Developer

JoshOrndorff:
This PR removes the distinction between Developer and Conceptual documentation as we discussed in the team meeting on 6 Jan.

While the table of contents is correctly updated, the file structure is not changed. I propose that we don't change the underlying file structure until we decide that we can abandon the v1.0 docs entirely. Because of a [bug in docusaurus](https://github.com/facebook/docusaurus/issues/845), files that existed in any previous version must still exist in the current version. Once we decide we can abandon the v1.0 docs, we can then clean up the file structure and cut a new version.

### DIFF
--- a/website/sidebars-full.json
+++ b/website/sidebars-full.json
@@ -1,123 +1,70 @@
 {
   "docs": {
-    "Overview": ["index", "getting-started", "glossary"],
-    "Conceptual": [
-      "conceptual/index",
-      {
-        "type": "subcategory",
-        "label": "Substrate Core",
-        "ids": [
-          "conceptual/core/codec",
-          "conceptual/core/consensus",
-          "conceptual/core/executor",
-          "conceptual/core/networking",
-          "conceptual/core/off-chain-workers",
-          "conceptual/core/storage"
-        ]
-      },
-      {
-        "type": "subcategory",
-        "label": "Substrate Nodes",
-        "ids": [
-          "conceptual/node/client",
-          "conceptual/node/light-client",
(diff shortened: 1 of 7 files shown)


website/sidebars-full.json
```diff
@@ -1,123 +1,70 @@
...
+    ],
+    "Substrate Core": [
+      "conceptual/core/codec",
+      "conceptual/core/consensus",
+      "conceptual/core/executor",
+      "conceptual/core/networking",
+      "conceptual/core/off-chain-workers",
+      "conceptual/core/storage"
+    ],
+    "Substrate Nodes": [
```
riusricardo: Why not call it `Substrate Client`. Then we can add the `client` part as an overview.
The section will explain the different types of clients, their parts and how to interact with them.

website/sidebars-full.json
```diff
@@ -1,123 +1,70 @@
...
-        ]
-      }
+    "Overview": [
+      "index",
+      "getting-started",
+      "conceptual/runtime/frame",
+      "glossary"
+    ],
+    "Substrate Core": [
+      "conceptual/core/codec",
```
riusricardo: Is it okay to keep codec in core? I think that it is more informative as a part to interact with the node/client. Even if all the data is encoded with it. Leaving it to you.

JoshOrndorff:
Closing this in favor of #406 which includes these changes. Let's take it all together.

@riusricardo Thanks for your review and suggestions. I'll incorporate them into #406

PR #404, opened 2020-01-06
Update consensus diagrams to Sveta's versions

JoshOrndorff:
Replaces the ugly hand-drawn consensus diagrams with beautiful new ones.

Screenshot of new diagrams


joepetrowski:
In the GHOST one, as I understand it, all numbers should be reduced by 1, i.e. all chain tips should be `0`. It's "blocks built on top of..." so you shouldn't count the block itself, no?

JoshOrndorff:
@joepetrowski Good catch. This was a mistake in my original diagram as well.

@goldsteinsveta Could you please reduce the numbers of each block in the GHOST diagram. Sorry for giving you bad source material.

goldsteinsveta:
@joepetrowski @JoshOrndorff I've reduced block count and image size in here

PR #415, opened 2020-01-10
Update seminar page

JoshOrndorff:
This PR makes several minor updates to the Seminar page:
* Updates youtube links to point to Parity's playlist rather than my own (at Zach's request)
* Includes last week's recording at the bottom of the page
* Includes upcoming topics in the FAQ


jimmychu0807:
I move the `col` div inside the `row` div in the previous recordings section.

I did a commit as I am thinking this is what you actually want. But feel free to revert if this is not the case

JoshOrndorff:
Yes, that's what I wanted. Thank you for catching it!

PR #420, opened 2020-01-15
Tutorial Reorganization

JoshOrndorff:
This PR attempts to add an even-lower-barrier-to-entry tutorial as discussed in the 13 Jan devex team meeting (specifically by @joepetrowski). It accomplished this by splitting the existing "Creating your first Substrate Chain" tutorial into two smaller tutorials.

Things changed
* Reduce scope of "Creating your first Substrate Chain" tutorial to just launching a node template and interacting via Apps
* Add "Build a dApp" tutorial which covers the PoE pallet and front end template
* Include both of these tutorials in the tutorial index
  * But doesn't add a catchy new logo to the dApp tutorial :-/
* Move substrate kitties tutorial to "Other Tutorials" section. (Not because it's bad, but because it's outdated, and we previously decided there should be 4 main tutorials)
* Remove requirement that each tutorial be self-contained from contributing guidelines.


jimmychu0807:
LGTM too. Minor update made.

PR #428, opened 2020-01-21
Fix front-end tutorial sidebar

jimmychu0807:
Sidebar for Front End Tutorial has disappeared previously. This fix adds them back.

### DIFF
--- a/website/sidebars.json
+++ b/website/sidebars.json
@@ -79,5 +79,14 @@
       "tutorials/start-a-private-network/keygen",
       "tutorials/start-a-private-network/customchain"
     ]
+  },
+  "substrate-front-end": {
+    "Substrate Front End Tutorial": [
+      "tutorials/substrate-front-end/index",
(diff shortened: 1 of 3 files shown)


docs/tutorials/substrate-front-end/index.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: "Introduction"
+id: substrate-front-end-index
```
JoshOrndorff: Is this necessary to make the sidebars come back?

jimmychu0807: By default the page has id `introduction`. But ID should be unique across all doc pages, and `introduction` as a doc page ID just seem too generic to me. So I specify one.

JoshOrndorff: Great. We should be careful of that in the future. Glad you caught it.

PR #432, opened 2020-01-22
Update getting-started

riusricardo:
Adds @kaichaosun's suggestion.

### DIFF
--- a/docs/overview/getting-started.md
+++ b/docs/overview/getting-started.md
@@ -18,6 +18,18 @@ Know that we are very happy to help you get started building with Substrate, so 
 
 To develop in the Substrate ecosystem, you must set up your developer environment. Depending on your operating system, these instructions may be different.
 
+## Fast Installation 
+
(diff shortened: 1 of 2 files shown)


docs/overview/getting-started.md
```diff
@@ -151,6 +161,17 @@ You can then run this generated binary with:
 subkey
 ```
 
+## Interact with Substrate
+
+The fastest way to interact with any local or public Substrate network is to visit Polkadot-JS Apps:
+
+[https://polkadot.js.org/apps/](https://polkadot.js.org/apps/)
```
JoshOrndorff:
```suggestion
[https://polkadot.js.org/apps/](https://polkadot.js.org/apps/rpc=ws://localhost:9944)
```

We can automatically point Apps at the local node. However, it is not trivial to interact with a locally running node from Apps because the local node will use an inseucre websocket endpoint.

PR #433, opened 2020-01-25
Clarify environment for private network tutorial

JoshOrndorff:
This PR addresses concerns raised in https://github.com/paritytech/substrate/issues/4735 by being more explicit that the learner should be using the single-machine environment described in the beginning of the tutorial. I'd like feedback not just about the technical correctness of the PR, but whether adding it makes the tutorial better.

Pro: If this is merged it is less likely that future learners will encounter the mismatched genesis block issue.

Con: Merging a PR like this every time someone gets stuck on a tutorial will eventually leave the tutorial covered with warnings. Too high a warning to instruction ratio is intimidating, and makes it challenging to follow the main thread of the tutorial.


riusricardo:
Don't forget to push the same changes to **pre-2.0**

PR #443, opened 2020-02-05
Use `--alice`/`--bob` cli flags instead -f `--key`

rakanalh:
I was following the tutorial and noticed the absense of `--key` from the node template... i updated the tutorial to use the correct flags.


rakanalh:
I just noticed that i updated the 1.0 version of the docs, which is what i was looking at on the website. This PR doesn't seem to be needed.

PR #445, opened 2020-02-06
Rename StorageMap::exists to contains_key

apopiak:
companion PR to https://github.com/paritytech/substrate/pull/4847

### DIFF
--- a/docs/tutorials/build-a-dapp/pallet.md
+++ b/docs/tutorials/build-a-dapp/pallet.md
@@ -175,7 +175,7 @@ decl_module! {
             let sender = ensure_signed(origin)?;
 
             // Verify that the specified proof has not been claimed yet or error with the message
-            ensure!(!Proofs::<T>::exists(&proof), "This proof has already been claimed.");
+            ensure!(!Proofs::<T>::contains_key(&proof), "This proof has already been claimed.");
 
             // Call the `system` pallet to get the current block number
             let current_block = <system::Module<T>>::block_number();
@@ -194,7 +194,7 @@ decl_module! {
             let sender = ensure_signed(origin)?;
 
             // Verify that the specified proof has been claimed
-            ensure!(Proofs::<T>::exists(&proof), "This proof has not been stored yet.");
+            ensure!(Proofs::<T>::contains_key(&proof), "This proof has not been stored yet.");
 
             // Get owner of the claim
             let (owner, _) = Proofs::<T>::get(&proof);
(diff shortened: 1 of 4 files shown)


apopiak:
Anything else I need to adjust? Or do we just wait?

riusricardo:
We could wait ~2 weeks to see if we have v2.0 at that point in time. Updating the node-template has a big impact on the DevHub and tutorials. We need to plan quite well every time we do it. It could impact even more this time because we restructured the node-template in master.

apopiak:
Ok, meaning we will come back to this if and when we update the `substrate-node-template`.
I'm guessing it does not make sense to open a PR for the node template right now as it will pull in quite a few changes.

riusricardo:
Correct, there is a [PR](https://github.com/substrate-developer-hub/substrate-node-template/pull/22) to update the template but it hasn't been merged because of the impact it will have. We need to be ready with PRs for all the docs an tutorials.

JoshOrndorff:
> Updating the node-template has a big impact on the DevHub and tutorials.

That's totally true and totally a pain in the butt. I recommend we solve this problem. Each time we cut a new version of the node template we tag it with something like `pre-v2.0-3e65111`. Then when we reference it from a tutorial, the tutorial should start with cloning node template and checking out the correct version of the template.

For example the [Creating Your First Substrate Chain tutorial](https://substrate.dev/docs/en/tutorials/creating-your-first-substrate-chain/setup#compiling-substrate) could be updated to read like this.
> Checkout the Substrate Node Template and create a branch for your work.
> ```bash
> git clone -b pre-v2.0-3e65111 https://github.com/substrate-developer-hub/substrate-node-template
> git checkout -b my-first-substrate-chain
> ```
>
> Initialize your WebAssembly build environment
> ...

Obviously that command would look less stupid if the tag were just `v2.0`, but such is life.

jimmychu0807:
Happy with Joshy suggestion that we also need to tag substrate-node-template, and then refer to the node-template tag in our documentations and tutorials. Just to clarify on the nomenclature, so it is 
`<some-version>-<substrate git rev 7 digits>`?

JoshOrndorff:
> Just to clarify on the nomenclature, so it is `<some-version>-<substrate git rev 7 digits>`?

That's the nomenclature for now until 2.0 is released. When Substrate 2.0 is released we will also tag the node template `v2.0`. After that we'll have to see how regular releases are. If upstream releases `v2.1` in a reasonable time, we'll follow them. If not we're back to `pre-v3.0-<hash>`.

riusricardo:
I agree, Joshy's suggestion should be the way to add tags to use in tutorials

JoshOrndorff:
With rustdoc built for pre-v2.0-3e65111, and tutorials now explicitly referencing the tag they use, I don't see any reason this can't be merged.

JoshOrndorff:
Subsumed by #493

PR #463, opened 2020-02-19
Forward all v1.0 getting started pages

shawntabrizi:
This is an example of how you would forward all of the v1.0 getting started pages to the correct page on the correct version.


shawntabrizi:
@riusricardo I was using Firefox too, and the behavior seems good to me. What is the alternative?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections#Alternative_way_of_specifying_redirections

riusricardo:
I got it from here: https://www.w3.org/TR/WCAG10-HTML-TECHS/#meta-element
>This element can specify metadata for a document including keywords, and information about the author. Please refer to the section on automatic page refresh for information on why META should not be used to redirect or auto-refresh pages.

>The following are deprecated HTML examples. The first changes the user's page at page at regular intervals. Content developers should not use this technique to simulate "push" technology. Developers cannot predict how much time a user will require to read a page; premature refresh can disorient users. Content developers should avoid periodic refresh and allow users to choose when they want the latest information.

I don't have an alternative. I believe you, it could be a specific be behavior of Firefox in Linux.

shawntabrizi:
Yeah I think the behavior is "discouraged" but I don't think it is deprecated anywhere:

https://stackoverflow.com/questions/5411538/redirect-from-an-html-page

I don't think it is trivial to do the redirect with JavaScript, which is I guess how it would be recommended today.

If you want to omit that logic, then a simple line of text with a link to the new page should suffice.

I leave that to you to decide.

JoshOrndorff:
Dumb question. Are these changes made to the `next` version (the files directly in the `docs` directory)? Shouldn't they be mode on the version 1.0 versioned docs?

riusricardo:
I think that Shawn's intention is to make it future proof. If you always point the link to `next` it should be safe.

shawntabrizi:
The point is if you go to:

```
https://substrate.dev/docs/en/1.0/getting-started/installing-substrate
```

You will still hit the `installing-substrate` page that you would expect. This is being permanently "cached" in the `versioned-docs/version-1.0` folder.

If you go to:

```
https://substrate.dev/docs/en/getting-started/installing-substrate

or

https://substrate.dev/docs/en/pre-2.0/getting-started/installing-substrate

or

https://substrate.dev/docs/en/next/getting-started/installing-substrate
```

You will get redirected, because the content which would populate all those pages is now a redirect.

shawntabrizi:
@riusricardo @JoshOrndorff @joepetrowski Is there something you are waiting on to merge here?

riusricardo:
Hey @shawntabrizi , I would like someone else to test you PR to see if they get the same issue I got.
If the meta issue happens to someone else, then I'll like to merge it with a warning/info message and remove the meta refresh/redirect.

riusricardo:
We could also remove the v1.0 docs but that will be an issue if we need to do it for v2.0 and then v3.0.

JoshOrndorff:
@riusricardo I checked out the branch, ran locally, and went to `localhost:300/docs/en/getting-started/installing-substrate`. I only got redirected a single time as expected. So it seems the redirect issue is pretty limited. I support merging.

riusricardo:
OK, thanks for this @JoshOrndorff :+1:

PR #471, opened 2020-02-26
Update visualizing-node-metrics.md

YessineAmor:
Added Prometheus configuration details and shared Grafana dashboard

### DIFF
--- a/docs/tutorials/visualizing-node-metrics.md
+++ b/docs/tutorials/visualizing-node-metrics.md
@@ -109,14 +109,33 @@ substrate_ready_transactions_number 0
 
 ## Next Steps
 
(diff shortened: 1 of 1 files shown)


docs/tutorials/visualizing-node-metrics.md
```diff
@@ -109,14 +109,33 @@ substrate_ready_transactions_number 0
 
 ## Next Steps
```
JoshOrndorff: The next steps heading should be retained at the bottom of the tutorial. `## Next Steps` is how we start to say "that's all this tutorial teaches, but here are some other readings you may enjoy"

YessineAmor: Got it. Thanks!

PR #485, opened 2020-03-05
Update Add a Pallet Tutorial

riusricardo:
Updates to 2.0-alpha.3+

Possible Enhancement:
- [ ] Use dependencies from crate.io instead of github  (Depends on crates.io status)


JoshOrndorff:
When we have an alpha.3 node template ready, we should update the instructions at the beginning to clone the proper tag so users start with the correct node template version. That's waiting on https://github.com/substrate-developer-hub/substrate-node-template/pull/25

PR #493, opened 2020-03-11
Updates three tutorials to alpha.3

JoshOrndorff:
Updates our first three tutorials (create first chain, build PoE dapp, start private network) to use alpha.3


docs/tutorials/build-a-dapp/prepare.md
```diff
@@ -26,7 +26,7 @@ Now you can proceed to set up the front-end template with these commands.
 
 ```bash
 # Clone the code from github
-git clone https://github.com/substrate-developer-hub/substrate-front-end-template
+git clone -b v2.0.0-alpha.3 https://github.com/substrate-developer-hub/substrate-front-end-template
```
riusricardo:
```suggestion
git clone -b v2.0.0-alpha.3 --depth 1 https://github.com/substrate-developer-hub/substrate-front-end-template
```

riusricardo: It may be good to add depth. The reason is for this is that at some point in time, the repo contained a zip file that added a few extra MBs.

JoshOrndorff: Is it true for the front end template too? If so, I agree we should add depth here. I thought it wasn't a problem for front end template though.

JoshOrndorff:
We shouldn't merge this until we've tagged the releases in the front end template. That is being addressed in https://github.com/substrate-developer-hub/substrate-front-end-template/pull/85

docs/tutorials/build-a-dapp/prepare.md
riusricardo: right, I didn't see that it was the front-end template. We don't need it there.

PR #499, opened 2020-03-18
Wrong filename in text

Swader:


JoshOrndorff:
This filename and the diagram above it are correct for the `v2.0.0-alpha.3` node template. They are referring to [this file](https://github.com/substrate-developer-hub/substrate-node-template/blob/v2.0.0-alpha.3/pallets/template/src/lib.rs).

If you're developing against an older node template (like `pre-v2.0-3e65111`) you should follow this version of the tutorial. https://substrate.dev/docs/en/tutorials/build-a-dapp/pallet

Swader:
Did I edit the wrong thing? https://substrate.dev/docs/en/next/tutorials/build-a-dapp/pallet - this tutorial has the wrong filename in text

PR #500, opened 2020-03-18
Remove DispatchResult

Swader:
Explicitly returning DispatchResult implies that one can change the result type of the functions. Rather than mislead the user this way, I propose removing DispatchResult and accompanying `Ok` returns, so that the macro takes care of that part for the user.

### DIFF
--- a/docs/tutorials/build-a-dapp/pallet.md
+++ b/docs/tutorials/build-a-dapp/pallet.md
@@ -101,7 +101,7 @@ Since imports are pretty boring, you can start by copying this at the top of you
 #![cfg_attr(not(feature = "std"), no_std)]
 
 use frame_support::{
-	decl_module, decl_storage, decl_event, decl_error, dispatch::DispatchResult, ensure, StorageMap
+	decl_module, decl_storage, decl_event, decl_error, ensure, StorageMap
 };
 use system::ensure_signed;
 use sp_std::vec::Vec;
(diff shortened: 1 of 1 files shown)


JoshOrndorff:
The more I think about this, the more I agree with @Swader . I'm okay with merging this PR, but want to be sure at least someone from the devhub team agrees first.

JoshOrndorff:
@riusricardo I'm also fine keeping `DispatchResult`. I'm not sure what you mean about errors macro though. This most recent version of the tutorial does have `decl_error!`.

riusricardo:
@JoshOrndorff  you're right, I didn't see the full code this time. Anyway, does the errors work if they don't return a DispatchResult?

JoshOrndorff:
Yeah, the errors still work even without an explicit `-> DispatchResult` because (I think) `decl_module!` adds that return type for you if you omit it.

It works with `ensure!` as well as `return Err(<Error<T>>::SomeCustomError.into());` although I would probably prefer to keep the explicit return type if you're using an explicit return statement (which we are not in this PoE tutorial).

joepetrowski:
To me, this depends on what we're trying to teach here. Most functions in FRAME do not explicitly return `DispatchResult`, but of course the macro adds it. Surely the knowledge base should have articles about runtime macros, but they are not very good at this time.

I like the idea of explaining the macros, at least at a high level, so it's not just "these are magic and trust us" but also keeping the code as simple as possible. I'll leave it to @JoshOrndorff to make the call, but IMO, I would:

1. Accept this change of removing `DispatchResult` from the _code_, and
2. Add a little text (~1 paragraph) saying that `decl_module` expands these functions to do a little bit more than you see.

My generalized opinion on stuff like this is: Showing code should always make an abstract idea concrete, but is never a shortcut to actually explaining how something works.

PR #501, opened 2020-03-18
Add doc explaining behaviour of ordered structures

pscott:
This PR aims to document the behavior of the codec when decoding structures.

Indeed, one might expert this test to work:
```rust
#[test]
fn binaryheap_decode_encode_test() {
    let mut bytes: &[u8] = &[8, 1, 0, 0, 0, 0, 0, 0, 1];
    let expected = bytes.clone();
    let obj = <BinaryHeap<u32>>::decode(&mut bytes).unwrap();
    let data: &[u8] = &obj.encode();
    assert_eq!(expected, data);
}
```

However this doesn't work: indeed when BinaryHeap decodes those bytes, it constructs a valid struct from them, which means that the structure inside of `obj` is actually represented this way `[8, 0, 0, 0, 1, 1, 0, 0, 0]`.

re: @kirushik

### DIFF
--- a/docs/conceptual/core/codec.md
+++ b/docs/conceptual/core/codec.md
@@ -133,7 +133,7 @@ Tuple of compact unsigned integer and boolean:
 
 `(3, false)`: `0x0c00`
 
-> For ordered structures, the decoder is expected to verify that incoming bytes are valid, and silently interpret bytes so that a valid ordered structure gets decoded. This implicitly means that decoding encoded bytes and then encoding this structure could result in having a encoded version that differs from the original bytes (e.g. if bytes are `[8, 1, 0, 0, 0, 0, 0, 0, 1]` and the expected structure is a `BinaryHeap<u32>` than decoding the original bytes and encoding the object will result in `[8, 0, 0, 0, 1, 1, 0, 0, 0]`).
+> For ordered structures, the decoder is expected to verify that incoming bytes are valid, and silently interpret bytes so that a valid ordered structure gets decoded. This implicitly means that decoding encoded bytes and then encoding this structure could result in having an encoded version that differs from the original bytes (e.g. if bytes are `[8, 1, 0, 0, 0, 0, 0, 0, 1]` and the expected structure is a `BinaryHeap<u32>` then decoding the original bytes and encoding the object will result in `[8, 0, 0, 0, 1, 1, 0, 0, 0]`).
 
 ### Enumerations (tagged-unions)
 
--- a/docs/conceptual/core/codec.md
+++ b/docs/conceptual/core/codec.md
@@ -120,12 +120,10 @@ SCALE Bytes:
 0x18040008000f00100017002a00
 ```
 
-### Tuples and Structures
+### Tuples
 
 A fixed-size series of values, each with a possibly different but predetermined
-and fixed type. This is simply the concatenation of each encoded value. For
-structures, the values are named, but that is irrelevant for the encoding (names
-are ignored - only order matters).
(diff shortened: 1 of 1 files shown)


JoshOrndorff:
> the decoder is expected to verify that incoming bytes are valid, and silently interpret bytes so that a valid ordered structure gets decoded.

What does this mean? What is the decoder supposed to do if they bytes are not valid? Are the bytes in your example case not valid?

pscott:
@JoshOrndorff So I had a hard time trying to explain this...
Basically, say the encoded bytes are invalid (for example `[8, 2]` is obviously invalid because there is not enough data to fill the buffer). Then, this should be an error.

The `invalid` I was referring to in this issue is the example I wrote in the PR: a BinaryHeap for example, stores the smallest element at the end of the vector, so the correct encoded form of a BinaryHeap should be `[8, 0, 0, 0, 1, 1, 0, 0, 0]`.
If the decoder receives `[8, 1, 0, 0, 0, 0, 0, 0, 1]`, he is expected to say "wait a second, this is not a valid BinaryHeap" (the smallest element is at the beginning of the vector). He is then expected to try and make a valid heap (or basically heapify the vector), in order to make a valid binaryheap out of it.
This is exactly what happens when we call `BinaryHeap::from()`.

Is that explanation clearer? How could I make the reader understand this concept in simpler terms?

JoshOrndorff:
Yes, this explanation is much clearer. Thank you!

If this note is worth adding, it's worth adding clearly, so don't sacrifice clarity for brevity. @joepetrowski is good at phrasing this kind of stuff.

bkchr:
That is way too complicated, we should just write that all containers store elements consecutively. The order of the elements is not fixed, depends on the container and can not be relied on at decoding.

docs/conceptual/core/codec.md
```diff
@@ -120,19 +120,37 @@ SCALE Bytes:
...
 
+### Data Structures
+
+For structures, the values are named, but that is irrelevant for the encoding (names are ignored - 
+only order matters). **All containers store elements consecutively. The order of the elements is not 
+fixed, depends on the container, and cannot be relied on at decoding.**
+
+The decoder is expected to verify that incoming bytes are valid, and silently interpret bytes so 
+that a valid structure gets decoded. This implicitly means that decoding some byte-array into a 
+specified structure and then re-encoding it could result in a different byte array than the
```
bkchr:
```suggestion
specified structure that enforces an order and then re-encoding it could result in a different byte array than the 
```

docs/conceptual/core/codec.md
```diff
@@ -120,19 +120,37 @@ SCALE Bytes:
...
 
 `(3, false)`: `0x0c00`
 
+### Data Structures
+
+For structures, the values are named, but that is irrelevant for the encoding (names are ignored - 
+only order matters). **All containers store elements consecutively. The order of the elements is not 
+fixed, depends on the container, and cannot be relied on at decoding.**
+
+The decoder is expected to verify that incoming bytes are valid, and silently interpret bytes so
```
bkchr: What you mean by "verify that incoming bytes are valid"?

pscott: I will just remove this confusing sentence regarding bytes validity

pscott:
> Sounds better, I think we should describe the general structure for containers:
> `Compact<u32> + [elements]`

Not sure I understand what you mean by that. Isn't it [described](https://substrate.dev/docs/en/conceptual/core/codec#vectors-lists-series-sets) already?

PR #504, opened 2020-03-19
Fixes Error metadata issue

Swader:


JoshOrndorff:
@Swader Did this solve your issue? It is indeed a correct change. But what a mystery that I didn't need it when I copied your code.

Swader:
@JoshOrndorff yes it solved it, it is a mystery indeed, it should not have worked for you

PR #511, opened 2020-03-25
Fix typos

riusricardo:


website/pages/en/tutorials.js
```diff
@@ -366,7 +366,7 @@ class Tutorials extends React.Component {
                 difficulty: "easy",
                 length: "2",
                 prerequisite: false,
-                version: "2.0-3e65111",
+                version: "2.0.0-alpha.3",
                 href: `${docUrl("tutorials/substrate-front-end/")}`
               },
```
JoshOrndorff: This tutorial has not actually been updated to alpha.3.

For example in this section, it explains checking a balance with `api.query.balances.freeBalance` which stopped working for alpha.3 https://substrate.dev/docs/en/tutorials/substrate-front-end/part-2-query-balances#22-query-and-display-alices-balance

JoshOrndorff: @Tbaut Do you want to maintain this tutorial? (I'm not pressuring you, I'm genuinely wondering. Any answer is fine.) Recently we've been maintaining the [Substrate front-end template](https://github.com/substrate-developer-hub/substrate-front-end-template/) which is originally based on your work in this tutorial.

website/pages/en/tutorials.js
```diff
@@ -286,7 +286,7 @@ class Tutorials extends React.Component {
                 difficulty: "easy",
                 length: "4",
                 prerequisite: false,
-                version: "2.0-3e65111",
+                version: "2.0.0-alpha.3",
```
JoshOrndorff: This tutorial isn't really versioned. See also https://github.com/substrate-developer-hub/substrate-contracts-workshop/issues/60

Up to you how to tag it here @riusricardo

riusricardo: RIght, it should be working with the latest tag. Good recommendation. Doing it that way, we can keep the version tag.

website/pages/en/tutorials.js
Tbaut: If you feel this tutorial is useful to anyone, I'm happy to help maintain it. I have been bumping deps regularly in the past and check that things kept working. Now if it's not viewed or used, I wouldn't mind removing it.

Do we have stats about it?

JoshOrndorff: We do have analytics.
image

These results could be better. That's probably partly because that tutorial hasn't been promoted as highly as others. IMO we need _some_ kind of tutorial on building a front end. It isn't realistic for everyone to use apps forever.

I'd like to do at least one of these:
1. Update that tutorial
2. Replace it with one that starts from the existing front end template
3. Both

Tbaut: ok, then let's update it.

JoshOrndorff: Moved this conversation to riot to not hold up this PR.

PR #512, opened 2020-03-25
Article on Debugging

JoshOrndorff:
This PR brings the beginning of an article on runtime debugging into the KB from the recipes. The work was started in substrate-developer-hub/recipes#134 and took a detour at https://github.com/substrate-developer-hub/knowledge-base/pull/9 on its way here.

### DIFF
--- a/docs/development/module/debugging.md
+++ b/docs/development/module/debugging.md
@@ -1,0 +1,146 @@
+---
+title: Debugging
+---
+
+Debugging is a necessary in all walks of software development, and blockchain is no exception. Most of the same tools used for general purpose Rust debugging also apply to Substrate. However, there are some restrictions when operating inside of a no_std environment like the Substrate runtime.
+
+
+## Substrate's Own `print` Function
+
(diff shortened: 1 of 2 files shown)


docs/development/module/debugging.md
```diff
@@ -0,0 +1,146 @@
...
+
+impl<T: Trait> Printable for Error<T> {
+	fn print(&self) {
+		match self {
+			Error::NoneValue => "Unexpected value".print(),
+			Error::StorageOverflow => "Value exceeded".print(),
+			_ => "Invalid case".print(),
+		}
+	}
+}
```
riusricardo: About this part, I was able to implement and build the code. The problem where I got stuck here was that it didn't print the text I wanted. It was printing the enum text as the error.

docs/development/module/debugging.md
```diff
@@ -0,0 +1,146 @@
...
+			Self::Module { index, error, message } => {
+				index.print();
+				error.print();
+				if let Some(msg) = message {
+					msg.print();
+				}
+			}
+		}
+	}
+}
```
riusricardo: This is the code from Substrate codebase where the conversion happens. Somehow it prefers to use the `error.print()` instead of the previous code where we implmented `impl<T: Trait> Printable for Error<T> `

PR #532, opened 2020-05-28
Move links to crates.io

danforbes:
This is a simple global find-and-replace. There should be a separate effort to verify all these links work, but I think this is at least an improvement.


danforbes:
Closing in favor of #533

PR #540, opened 2020-06-04
Fix links and HTML comments

riusricardo:


docs/knowledgebase/runtime/metadata.md
```diff
@@ -456,8 +456,6 @@ the FRAME System pallet is declaring that it may raise the
 
 <!--
 ### Examples
-
--
```
JoshOrndorff, line 460: @riusricardo So having blank lines inside HTML comments breaks the page?

riusricardo: Not sure if it is the case for everything. I found out that it happens with the HTML comments. It breaks the TOC on the right.

PR #559, opened 2020-06-09
Enable metrics tutorial

riusricardo:
Enables the node metrics tutorial


riusricardo:
It is marking `http://localhost:3000/docs/tutorials/create-your-first-substrate-chain` as broken link. My guess is that it is not resolving the link because it misses the language.
Docusaurus resolves it correctly in the web-browser.
`http://localhost:3000/docs/en/tutorials/create-your-first-substrate-chain`

PR #565, opened 2020-06-12
Update Knowledge Base FRAME Documentation

danforbes:
Closes https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/566

### DIFF
--- a/docs/knowledgebase/runtime/frame.md
+++ b/docs/knowledgebase/runtime/frame.md
@@ -70,6 +70,13 @@ included with your runtime and configures them to work together to compose your 
 calls are made to your runtime, it uses the Executive pallet to dispatch those calls to the
 individual pallets.
 
+### Benchmarking
+
+Macro for benchmarking a FRAME runtime.
+
+- [Docs](https://crates.parity.io/frame_benchmarking/index.html)
+- [Source](https://github.com/paritytech/substrate/blob/master/frame/benchmarking/src/lib.rs)
+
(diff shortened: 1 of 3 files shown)


burdges:
Add some note that Randomness Collective Flip exists for tests and should not be used in production.

JoshOrndorff:
This seems like a good place to change the text "executive pallet" -> "frame executive". According ti [its own docs](https://crates.parity.io/frame_executive/index.html), "The executive module is not a typical pallet providing functionality around a specific feature". I would argue that it is not a pallet at all because:

* It doesn't use `decl_module!`
* it isn't included in `construct_runtime!`
* It doesn't have a configuration trait

JoshOrndorff:
Another thought: This article was initially written at a time when the pallets in the substrate repo were basically the only useful pallets. That's not the case anymore.

I think this article should explain that FRAME is a technique for writing runtimes that uses the support libraries, the execute, etc. I think it's still fine to discuss the pallets in the Substrate repo, but it should be clear that anyone can write a pallet, and maybe mention some good additional sources of pallets (polkadot runtime, [orml](https://github.com/open-web3-stack/open-runtime-module-library/), recipes, substrate marketplace)

danforbes:
@wheresaddie - from my POV the statement that a module is intended for testing/benchmarking purposes and not for use in production is somewhat self-evident. I'm happy to hear other's thoughts on this topic or specific suggestions for how the verbiage can be improved.

Issue #556, opened 2020-06-08
Update 404 page


shawntabrizi:
I see a lot of comments like:

```
I can't find it on the website where it used to be https://substrate.dev/docs/en/next/tutorials/visualizing-node-metrics#step-3-configure-prometheus-to-scrape-your-substrate-node
```

based on the latest updates to the developer hub. I suggest you do a simple update to your 404 page so that people who land on these old links are automatically redirected to the right place.

wheresaddie:
updating now. thanks

PR #567, opened 2020-06-12
Redirect Codec page

JoshOrndorff:
This PR adds a redirect from a previous location of the scale codec doc to the new one. Specifically it adds a redirect from `substrate.dev/docs/en/conceptual/core/codec` to `substrate.dev/docs/en/knowledgebase/advanced/codec`.

There are several weaknesses of this approach to redirects.
1. It requires creating a file in our directory for each old location that we want to redirect.
2. It does _not_ redirect the other old location for this doc `substrate.dev/docs/en/next/conceptual/core/codec

If it is possible it would be preferable to implement the redirect at the DNS / cloudflare level. cc @lovelaced

### DIFF
--- a/docs/conceptual/core/codec.md
+++ b/docs/conceptual/core/codec.md
@@ -1,0 +1,3 @@
+<meta http-equiv="Refresh" content="0; url=../../knowledgebase/advanced/codec" />
+
+View the latest docs on the [SCALE codec](../../knowledgebase/advanced/codec).


JoshOrndorff:
@shawntabrizi suggested another alternative implementation would be to add javascript to our 404 page that detects these old urls and redirects to the correct location based on a hard-coded mapping.

ref https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/556

JoshOrndorff:
Closing in favor of #569 which takes Shawn's approach

danforbes:
Should we close this one, Joshy?

Issue #39, opened 2019-02-27
System Module


shawntabrizi:
Custom Documentation:

- [ ] The purpose of the module
- [ ] Explanation of the real world users and their role/interactions with the module
- [ ] Key logic, processes, or assumptions that happen within the module
- [ ] Details on how a module easily integrates with other associated modules

Rust Docs:

- [ ] Available storage items and what they hold
- [ ] Public Functions
- [ ] Internal Functions

gautamdhameja:
PR https://github.com/paritytech/substrate/pull/1999 merged. Closing this issue.

PR #573, opened 2020-06-15
Enable Contracts tutorial

riusricardo:
Enables contracts tutorials again.
Closing [tutorials #39](https://github.com/substrate-developer-hub/tutorials/issues/39) now that the UI has been fixed.

### DIFF
--- a/website/pages/en/tutorials.js
+++ b/website/pages/en/tutorials.js
@@ -291,6 +291,22 @@ class Tutorials extends React.Component {
                 )}`
               },
               {
+                img: `${baseUrl}img/ink-smart-contracts-tutorial.png`,
+                title: <translate>ink! Smart Contracts Tutorial</translate>,
+                text: (
+                  <translate>
+                    A comprehensive, end-to-end tutorial for building an ERC20
+                    token contract using ink!.
+                  </translate>
+                ),
+                difficulty: "easy",
+                length: "4",
+                prerequisite: false,
+                version: "2.0.0-rc4",
+                href:
+                  "https://substrate-developer-hub.github.io/substrate-contracts-workshop/"
+              },
+              {
                 img: `${baseUrl}img/crates.png`,
                 title: (
                   <translate>Write a Pallet in its Own Crate</translate>
--- a/docs/knowledgebase/integrate/memory-profiling.md
+++ b/docs/knowledgebase/integrate/memory-profiling.md
@@ -1,0 +1,97 @@
+---
+title: Memory Profiling
+---
+
+Memory profiling enables you to understand the memory allocation and behavior of your blockchain
+applications over time in Substrate-based clients. It identifies method calls in the context of how
+memory was allocated, combining this information with the number of allocated objects. In addition,
+profiling can be used to analyze memory leaks, identify where memory consumption is happening,
+define temporary allocations, and investigate excessive memory fragmentation within applications.
+When running Substrate locally, it also accounts for runtime.
+
+## Understanding Modes
+
+## Installing memory-profiler
+
+We recommend [koute's memory profiler](https://github.com/koute/memory-profiler). This assumes you
+have installed rust stable, GCC, and yarn globally.
+
+[The following steps will build a memory profiler](https://github.com/koute/memory-profiler#building):
+
+1. clone the repository
+2. compile the library and the cli
+
+```bash
+git clone https://github.com/koute/memory-profiler
+cd memory-profiler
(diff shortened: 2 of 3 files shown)


website/pages/en/tutorials.js
```diff
@@ -290,6 +290,22 @@ class Tutorials extends React.Component {
...
+                text: (
+                  <translate>
+                    A comprehensive, end-to-end tutorial for building an ERC20
+                    token contract using ink!.
+                  </translate>
+                ),
+                difficulty: "easy",
+                length: "4",
+                prerequisite: false,
+                version: "2.0.0-dev",
```
danforbes, line 305: I don't think we should introduce a new, ad hoc version.

danforbes: To this end, did you actually go through this tutorial step-by-step to ensure that it works with `rc3` and the screenshots, etc, are at least mostly up-to-date?

riusricardo: I don't see any problem with this new tag. The problem is that many times this tutorial moves at different speed.
I tried at the rc3 tag but it doesn't works due to a small bug which is fixed in master. THis is why I use `dev` instead of `rc3`

danforbes: I don't want to enable a tutorial if it doesn't work on a released version.

JoshOrndorff: This is the exact problem I had in recipes that made me start experimenting with gitflow. The classic situation is:

There is a bug in Substrate's latest release (currently rc3)  that prevents a devhub product from working (e.g. contracts workshop). The bug is fixed on master and the devhub product works when using master. But because we want to stick to releases only, we cannot publish until another release is cut.

Don't have a concrete solution, but it's worth noticing this is not an isolated case. This will also affect other teams building on Substrate who might say, "Our chain has a bug because Substrate has a bug. We'd like to release a new node that contains the fix, but Substrate itself hasn't released yet, so we're stuck."

danforbes: I think this is a good time to go by the "three versions back" rule. If there is a released version within the last three versions, we can keep the tutorial up...otherwise, it doesn't seem like the feature is stable/important enough to maintain a tutorial for it.

riusricardo: A problem with teh last three versions is what I had to do in the past 2-3 months for this tutorial until it had many patches. It was using alpha.6 because it worked and also the UI. THe alpha.6 needed an older nightly to build (1st patch). The second problem came when the UI followed more breaking changes from master and I added another patch to tell the users about the data types. At the end the tutorial was pretty broken because you had to patch rustc and the UI.

wheresaddie: seems like it would be worth considering gitflow for devhub as this isn't an isolated case-- a lot of feedback loops dependent on bug patches --

wheresaddie: > I think this is a good time to go by the "three versions back" rule. If there is a released version within the last three versions, we can keep the tutorial up...otherwise, it doesn't seem like the feature is stable/important enough to maintain a tutorial for it.

yeap this-- maybe we should consider a /b/ side to devhub or sandbox where we can play with tutorials that require more experimental, less stable, builders in mind-- or people who are advanced enough they are willing to debug and patch broken rust/UI

riusricardo:
>Let's make sure this works with rc3 and update the version to reflect that. This is a nice looking tutorial and it will be a big addition to our tutorials rocket In fact, I think this may be a nice candidate UI to use for improved tutorials moving forward. This looks like the same UI Kitties is using, for instance.

@danforbes there is a catch here :smile:  The node-cli binary is on version `rc3` but at a different commit, so it is valid to say that is `2.0.0-rc3` as it is what it displays when you execute it.
https://github.com/paritytech/substrate/blob/master/bin/node/cli/Cargo.toml#L3
This tutorial doesn't requires node-template.

danforbes:
That implies that this tutorial is pinned to `master`, which means it could break anytime. I feel pretty strongly that we should only maintain tutorials for recently released versions.

riusricardo:
Lets just wait until the next release. I'll keep this open.

wheresaddie:
> Lets just wait until the next release. I'll keep this open.

makes total sense.. good catch @danforbes

danforbes:
These changes were pulled into https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/588 - many thanks, @riusricardo :pray:

PR #574, opened 2020-06-15
Custom redirect for broken pages

shawntabrizi:
This adds a small JS script to every page of the developer hub.

We check the current url, and see if any of the desired redirect patterns match.

If so, we take the user to the new page, leaving the old page in history.

This script currently replaces instances of `docs/en/next` or `docs/next` with `docs/en/knolwedgebase` or `docs/knolwedgebase` which should handle the majority situation.

Custom full redirects beyond that can be added given the template I have provided, or this script can be modified for the needs of the site.

### DIFF
--- a/website/pages/en/404.js
+++ b/website/pages/en/404.js
@@ -34,7 +34,7 @@ class ErrorPage extends React.Component {
           !
         </p>
          <p>
-       Alternatively, if you arrive here via a dead link, look at the most recent <a href="https://github.com/paritytech/substrate">documentation</a> and contributions.
+       Alternatively, if you arrive here via a dead link, return to the <a href="https://substrate.dev/docs/">documentation homepage</a>.
       </p>
         <p>
           <Button href="/" variant="secondary" className="primary-color">
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -88,6 +88,7 @@ const siteConfig = {
 		'/js/clipboard.min.js',
 		'/js/code-block-buttons.js',
 		'/js/load.js',
+		'/js/redirect-next.js',
 		{
 			src: '/js/ui.js',
 			defer: true
--- a/website/static/js/redirect-next.js
+++ b/website/static/js/redirect-next.js
@@ -1,0 +1,33 @@
+window.addEventListener("load", function() {
+    let pathname = window.location.pathname;
+
+    // Full, hardcoded redirects based on current path
+    let redirects = {
+        // "/impossible/example/page/": "/new/page/here/"
+    };
+    // Custom full redirects first
+    if (pathname in redirects) {
+        window.location.href = redirects[pathname];
+        return;
+    }
+
+    // We want to redirect:
+    // * https://substrate.dev/docs/next/...
+    // ------------------------^ [1] ^[2]
+    // * https://substrate.dev/docs/en/next/...
+    // ------------------------^ [1] --^ [3]
+    // pathArray[0] is an empty string
+
+    var pathArray = pathname.split('/');
+
+    // General redirect of user from `next` to `knowledgebase`
+    if (pathArray[1] == 'docs') {
+        if (pathArray[2] == 'next') {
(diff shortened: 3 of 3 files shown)


shawntabrizi:
This probably needs to be customized further for all the ways that the dev hub has changed before being merged. Feel free to modify this PR and merge it at your own leisure.

joepetrowski:
> which should handle the majority situation.

Actually I imagine this will break about 80% of the time. When I transitioned from here to the `knowledge-base` repo I reorganized the directory structure. We don't even have these `conceptual` and `developer` prefixes anymore.

shawntabrizi:
@joepetrowski then you should customize the logic so that it handles your restructuring. In general I am not very aware of what was done.

wheresaddie:
makes sense to do a sweeping check for broken pages/links and either update them or nuke them- alternatively or in addition our 404 might reflect commonly searched for topics/spaces as a static list, including a reference to the riot chat for additional support

joepetrowski:
> makes sense to do a sweeping check for broken pages/links and either update them or nuke them- alternatively or in addition our 404 might reflect commonly searched for topics/spaces as a static list, including a reference to the riot chat for additional support

Yes, I am 100% in favor of a 404 page with more info and a note that we've restructured. _There really is no pattern to match._ The old DevHub was deep in technical debt and I manually went through the articles one by one and either moved them to a new folder or deleted them (I found 3 "intro to ink" pages in different directories under the old structure...).

I know this is pessimistic but I think there are two options:

1. Redirect a few pages and show a useful error message for the rest.
2. Look at this commit and hardcode all these changes: https://github.com/substrate-developer-hub/knowledgebase/commit/4dae04260f3069150d488395c6c45c516c334ad1

riusricardo:
@danforbes I would say that this would be an improvement. It's a small one but it could be complemented by other type of redirections. I think that it won't really break what exists, it would be another step to catch an incorrect url in case that they get through the first redirect.

wheresaddie:
> @danforbes I would say that this is an improvement. It's a small one but it could be complemented with other type of redirections. I think that it won't really break what exists, it would be another step to catch an incorrect url in case that it gets through the first redirect.

So Id suggest we push a 404 page with more info and a note that we've restructured and are reworking the dev hub-- if people have additional issues suggestion they go to riot and write up a quick issue for us

danforbes:
That works for me. I pulled down this branch and tried to access the redirect logic and couldn't get anything useful to happen. I asked @riusricardo to look into things on his end to see if I'm just missing something. I am all for a more descriptive and helpful 404 page.

riusricardo:
The logic is correct but Docusaurus ignores it.
I ran the script steps in the console and it does the redirect. But if I try to access the url directly or I modify it, it doesn't work.

shawntabrizi:
The script works, you just need to set up your computer to serve the pages that we want to redirect from. You can test by creating dummy pages under a `next` folder

riusricardo:
> The script works, you just need to set up your computer to serve the pages that we want to redirect from. You can test by creating dummy pages under a `next` folder

If I understand correctly, what you are saying is doing something similar to the old redirect approach.
We need to recreate the old structure of docs with `empty files` under a `next` folder just to make this work.
Don't take it wrong, but from the analytics I saw that most of the invalid landing pages weren't using next or a version. However google could still have old docs in their cache.
IMHO looks to messy for a possible <20% success rate. If everyone agrees to this, we can do it.

shawntabrizi:
@riusricardo you are incorrect here.

You only need to do this to simulate the redirect locally. The behavior of your local server versus the actual hosted server is not the same. On a local server, if you go to a page that does not exist you are not served any page. You just get `Cannot GET`.

If you go to a missing page on the real server, you are hosted a `404` page and hence you are also hosted this script.

riusricardo:
Thanks for the clarification @shawntabrizi.
I'm good with the changes. We should commit and test. Then a follow-up PR could expand it.

Issue #578, opened 2020-06-19
A guide to on-chain randomness


gnunicorn:
- understanding the coin flip
 - understanding the outer assumptions one can and can't take from the traits
 - hardening the randomness, which are useful for what scenario, which drawbacks comes with them

PR #579, opened 2020-06-19
Guide to on-chain randomness

JoshOrndorff:
This PR addresses #578 but explaining:

* The problem of on-chain randomness
* Substrate's `Randomness` trait
* A few implementations of the Randomness trait
* Security concerns

It may also explain ways to strengthen the security if that is advisable. @gnunicorn You mentioned some ways to "strengthen" the randomness in #578. Are you confident in their security to add them here?


docs/knowledgebase/runtime/randomness.md
```diff
@@ -0,0 +1,92 @@
...
+be provided. The `Randomness` Trait provides two methods for gaining randomness.
+
+The first method is called `random_seed`. It takes no parameters and gives back a raw piece of
+randomness. Calling this method multiple times in a block will return the same value each time.
+Thus, it is not recommended to use this method directly.
+
+The second method is called `random`. It takes a byte-array which is used as a context identifier.
+It returns a result that is influenced by both the raw random seed and the context provided. This
+method may be called multiple times in the same block and will return a result that is specific to
+the context provided and independently low-influence from any other context.
```
burdges, line 46: Appears you got this from https://crates.parity.io/frame_support/traits/trait.Randomness.html#foreign-impls and it's true pallet-randomness-collective-flip should use the context prior to its low-influence function, which gives some independence, but pallet-randomness-collective-flip should be avoided outside of tests.  I'm dubious that "independently low-influence" makes sense, so..

"The second method is called `random`.  It takes a byte-array which is used as a context identifier and returns a result as unique to this context and as independently from other contexts as allowed by the underlying randomness source."

burdges:
Again, I suggest the ".. as independent as allowed by .." language because one should not imply that these actually satisfy some independence properties with randomness collective flip so easy to use, i.e. missuse.

JoshOrndorff:
:pray: Dear link-checking gods, please let the CI pass :pray:

PR #590, opened 2020-06-28
Minor Overhaul

danforbes:
I would suggest that the best way to review this is to take a quick look at the diff & then pull it down and run it locally.

- Rustdocs should point to substrate.dev/rustdocs
- Minor structural changes to reinforce latest DevHub POV

### DIFF
--- a/docs/knowledgebase/advanced/consensus.md
+++ b/docs/knowledgebase/advanced/consensus.md
@@ -50,7 +50,7 @@ parent. Forks must be resolved such that only one, canonical chain exists.
 
 A fork choice rule is an algorithm that takes a blockchain and selects the "best" chain, and thus
 the one that should be extended. Substrate exposes this concept through the
-[`SelectChain` Trait](https://crates.parity.io/sp_consensus/trait.SelectChain.html).
+[`SelectChain` Trait](https://substrate.dev/rustdocs/v2.0.0-rc4/sp_consensus/trait.SelectChain.html).
 
 Substrate allows you to write a custom fork choice rule, or use one that comes out of the box. For
 example:
@@ -59,7 +59,7 @@ example:
 
 The longest chain rule simply says that the best chain is the longest chain. Substrate provides this
 chain selection rule with the
-[`LongestChain` struct](https://crates.parity.io/sc_consensus/struct.LongestChain.html).
+[`LongestChain` struct](https://substrate.dev/rustdocs/v2.0.0-rc4/sc_consensus/struct.LongestChain.html).
 GRANDPA uses the longest chain rule for voting.
 
 ![longest chain rule](assets/consensus-longest-chain.png)
@@ -129,12 +129,12 @@ Developers are always welcome to provide their own custom consensus algorithms.
 
 ### Aura
 
-[Aura](https://crates.parity.io/sc_consensus_aura/index.html) provides a slot-based
+[Aura](https://substrate.dev/rustdocs/v2.0.0-rc4/sc_consensus_aura/index.html) provides a slot-based
 block authoring mechanism. In Aura a known set of authorities take turns producing blocks.
 
 ### BABE
 
-[BABE](https://crates.parity.io/sc_consensus_babe/index.html) also provides slot-based
+[BABE](https://substrate.dev/rustdocs/v2.0.0-rc4/sc_consensus_babe/index.html) also provides slot-based
 block authoring with a known set of validators. In these ways it is similar to Aura. Unlike Aura,
 slot assignment is based on the evaluation of a Verifiable Random Function (VRF). Each validator is
 assigned a weight for an _epoch._ This epoch is broken up into slots and the validator evaluates its
@@ -149,7 +149,7 @@ in a given slot. These "secondary" slot assignments allow BABE to achieve a cons
 
 ### Proof of Work
 
-[Proof-of-work](https://crates.parity.io/sc_consensus_pow/index.html) block authoring
+[Proof-of-work](https://substrate.dev/rustdocs/v2.0.0-rc4/sc_consensus_pow/index.html) block authoring
 is not slot-based and does not require a known authority set. In proof of work, anyone can produce a
 block at any time, so long as they can solve a computationally challenging problem (typically a hash
 preimage search). The difficulty of this problem can be tuned to provide a statistical target block
@@ -157,7 +157,7 @@ time.
 
 ### GRANDPA
 
-[GRANDPA](https://crates.parity.io/sc_finality_grandpa/index.html) provides block
+[GRANDPA](https://substrate.dev/rustdocs/v2.0.0-rc4/sc_finality_grandpa/index.html) provides block
 finalization. It has a known weighted authority set like BABE. However, GRANDPA does not author
 blocks; it just listens to gossip about blocks that have been produced by some authoring engine like
 the three discussed above. GRANDPA validators vote on _chains,_ not _blocks,_ i.e. they vote on a
@@ -173,7 +173,7 @@ coordination with the runtime. Examples include adjustable difficulty in proof o
 rotation in proof of authority, and stake-based weighting in proof-of-stake networks.
 
 To accommodate these consensus features, Substrate has the concept of a
-[`DigestItem`](https://crates.parity.io/sp_runtime/enum.DigestItem.html), a message
+[`DigestItem`](https://substrate.dev/rustdocs/v2.0.0-rc4/sp_runtime/enum.DigestItem.html), a message
 passed from the outer part of the node, where consensus lives, to the runtime, or vice versa.
 
 ## Learn More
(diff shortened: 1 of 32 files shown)


docs/knowledgebase/advanced/consensus.md
```diff
@@ -4,14 +4,17 @@ title: Consensus
...
 ## State Machines and Conflicts
 
 A blockchain runtime is a [state machine](https://en.wikipedia.org/wiki/Finite-state_machine). It
 has some internal state, and state transition function that allows it to transition from its current
 state to a future state. In most runtimes there are states that have valid transitions to multiple
-future states, but a single transition must be selected.
+future states, but a single transition must be selected. In some state machine models, it is also possible to associate actions with a state:
+
+- an entry action: performed when entering the state, and
+- an exit action: performed when exiting the state.
```
JoshOrndorff, line 17: I'm personally interested in this because I think a lot about state machines. Have a link I could read?

When I originally wrote this article I had a whole section about conflict exclusion before getting to forks, but we agreed to remove it because it was not relevant to Substrate which considers any fork a conflict even if the blocks could commute. I feel the same about this comment regarding alternate state machines.

docs/knowledgebase/advanced/consensus.md
```diff
@@ -20,7 +23,7 @@ Blockchains must agree on:
 - A final (current) state.
 
 In order to agree on the resulting state after a transition, all operations within a blockchain's
-[state transition function](../runtime/index) must be deterministic.
+[state transition function](../runtime/index) must be deterministic. In order to be deterministic, the same operation performed across different nodes should return the same result. A difference in results between the nodes for the same operation can output a failure in consensus as storing data on the ledger will lead to an inconsistent ledger state, thereby making the whole smart contract unusable.
```
JoshOrndorff, line 26: > can output a failure in consensus as storing data on the ledger will lead to an inconsistent ledger state, thereby making the whole smart contract unusable

I don't really understand the phrase "can output a failure in consensus" and I don't see why smart contracts are relevant here

docs/knowledgebase/advanced/consensus.md
```diff
@@ -31,7 +34,7 @@ different orders, and thus they must use a more elaborate method to exclude tran
 further complication, blockchain networks strive to be fault tolerant, which means that they should
 continue to provide consistent data even if some participants are not following the rules.
 
-Blockchains batch transactions into blocks and have some method to select which participant has the
+Blockchains batch transactions into blocks with methodologies to select which participant has the
```
JoshOrndorff, line 37: Why plural? Each chain has a single method to elect the next block author. And why switch to "with"? Before being modified, this sentence made two statements about blockchains. First, that they batch transactions into blocks, and second that they have a method to select... The modified version sounds to me like the "methodologies" ((probably meaning the consensus engine) somehow perform the batching.

docs/knowledgebase/advanced/consensus.md
```diff
@@ -71,18 +74,17 @@ resolved by choosing the branch that has the most blocks built on it recursively
...
 ![GHOST rule](assets/consensus-ghost.png)
 
-## Block Production
+## Block Authoring
 
 Some nodes in a blockchain network are able to produce new blocks, a process known as authoring.
 Exactly which nodes may author blocks depends on which consensus engine you're using. In a
 centralized network, a single node might author all the blocks, whereas in a completely
-permissionless network, an algorithm must select the block author at each height.
+permissionless network, an algorithm must select the block author at each height. This iterative process confirms the integrity of the previous block.
```
JoshOrndorff, line 82: I don't think the addition is correct. The order is.
1. The node confirms the integrity of the previous block by re-executing it
2. The node runs the consensus algorithm, and if it is selected to author, it authors on top of the best block it confirmed to be valid.

Thinking about it more now, maybe you meant something like, "By authoring a block, the author attests that the parent block and all of its ancestors recursively were valid"?

docs/knowledgebase/advanced/consensus.md
```diff
@@ -71,18 +74,17 @@ resolved by choosing the branch that has the most blocks built on it recursively
...
 
 ### Proof of Work
 
-In proof-of-work systems like Bitcoin, any node may produce a block at any time, so long as it has
-solved a computationally-intensive problem. Solving the problem takes CPU time, and thus miners can
-only produce blocks in proportion with their computing resources. Substrate provides a proof-of-work
+In proof-of-work (PoW) systems like Bitcoin, any node may produce a block at any time, so long as it has
+solved a computationally-intensive problem, this creates a consensus mechanism. By utilizing proof-of-work, the node deters denial-of-service attacks and other service abuses by requiring some work from the service requester, usually this translates into processer time. Solving the problems take CPU time, and thus miners only produce blocks in proportion with their computing resources. Substrate provides a proof-of-work
```
JoshOrndorff, line 87: First , there is a comma splice where you first started editing.

Second, the PoW is for the right to author a block, not the right to query a node. The PoW here doesn't provide DoS protection. It differs from hashcash in this way.

docs/knowledgebase/advanced/cryptography.md
```diff
@@ -6,10 +6,10 @@ This document offers a conceptual overview of the cryptography used in Substrate
 
 ## Hashing Algorithms
 
-Hash functions are used in Substrate to map arbitrary sized data to fixed-sized values.
+Hash functions are used in Substrate to map arbitrary sized data to fixed-sized values.  An algorithmically generated hash is essential to blockchain management in cryptocurrency.
```
JoshOrndorff, line 9: What other types of hash are there?

docs/knowledgebase/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "knowledgebase",
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "Unlicense",
+  "scripts": {
+    "update-doc-versions": "find . -type f -name \"*.md\" -print0 | xargs -0 sed -i -e 's/substrate.dev\\/rustdocs\\/'\"$OLD_VERSION\"'/substrate.dev\\/rustdocs\\/'\"$NEW_VERSION\"'/g'"
+  }
+}
```
JoshOrndorff, line 9: Can you say a  little about this package.json file?

Why is it nested inside the repo? Is this update script supposed to work alongside docusaurus's own versioning? https://docusaurus.io/docs/en/versioning

danforbes: This could easily be moved elsewhere, I was just looking for a place to encapsulate the logic of updating all the Rustdocs links when a new version of Substrate is released. I would say this is intended to be used _as opposed to_ Docusaurus's own versioning. Here's how this little script can be used to update the links: `OLD_VERSION=v2.0.0-rc3 NEW_VERSION=v2.0.0-rc4 yarn update-doc-versions`

danforbes:
@JoshOrndorff - I don't really think this PR makes the jump of versioning the Knowledge Base, so to speak. The Knowledge Base will still be on a floating version ("the latest released version"), which is just a _little_ more stable than pinning to `master`. I think in general we should think pretty specifically about only supporting released versions of Substrate.

docs/knowledgebase/advanced/consensus.md
joepetrowski: > Thinking about it more now, maybe you meant something like, "By authoring a block, the author attests that the parent block and all of its ancestors recursively were valid"?

Agreed. I don't think any new sentence is necessary here, but the "iterative process" one has to go.

docs/knowledgebase/advanced/consensus.md
joepetrowski: Yeah, revert this change. Besides what Joshy said, PoW does not create a consensus mechanism.

docs/knowledgebase/advanced/cryptography.md
```diff
@@ -35,11 +35,11 @@ Substrate uses the [`blake2`](https://docs.rs/blake2/) implementation in Rust.
 
 ## Public-Key Cryptography
 
-Public-key cryptography is used in Substrate to provide a robust authentication system.
+Public-key cryptography is used in Substrate to provide a robust authentication system. The application of a public key is commonly utilized in encrypting communication to provide confidentiality by obfuscating the recipient's public key which can only be decrypted by the recipient's paired private key.
```
joepetrowski, line 38: Revert, has nothing to do with encryption.

docs/knowledgebase/advanced/consensus.md
joepetrowski: What does this addition mean? Entry and exit actions have nothing to do with the state.

docs/knowledgebase/advanced/consensus.md
joepetrowski: Indeed, a consensus failure is not equivalent to conflicting opinions on state. In fact, consensus is made to resolve conflicting opinions on state.

docs/knowledgebase/advanced/cryptography.md
wheresaddie: < forehead slap >
</forehead slap>

docs/knowledgebase/advanced/cryptography.md
wheresaddie: correct that should of read private.. obviously I will never edit anything prior to coffee again

PR #601, opened 2020-07-21
Add grandpa blocks finalization comment

riusricardo:
Add comment to custom chain about Grandpa blocks finalization.


danforbes:
I'm not totally sure I think this comment is necessary. I think in the context of this tutorial it may actually make things more confusing. I feel like this tutorial makes it pretty clear that block finalization will not happen until both validators have joined the network. I think the "two-thirds" point may be an unnecessary implementation detail at this level of abstraction.

riusricardo:
I think that the comment is needed. You also said that it was confusing for you the actual text. The "two-thirds" point makes it super clear and is part of grandpa's design.

PR #608, opened 2020-07-28
Replaces images that use outdated SRML terminology

danforbes:


danforbes:
Preview number 1
image

Preview number 2
image

wheresaddie:
Looks great @danforbes ! I'm not big on the 'usage' diagram as I find it having two x-axis's to be visually confusing. I might suggest we take that out and just write one sentence or so for each dev track.. but that's probably a different PR :+1:

danforbes:
Shared your comment with Zsofia, @wheresaddie...thanks you for sharing. As you  know I will take all the design assistance I can get :sweat_smile:

PR #614, opened 2020-08-03
Update subkey.md

wheresaddie:
adding explanation of mnemonic phrase/dev phrase deployment

### DIFF
--- a/docs/knowledgebase/integrate/subkey.md
+++ b/docs/knowledgebase/integrate/subkey.md
@@ -322,7 +322,7 @@ Secret Key URI `//Alice` is account:
   SS58 Address:     5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
 ```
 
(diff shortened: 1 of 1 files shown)


docs/knowledgebase/integrate/subkey.md
```diff
@@ -322,17 +322,11 @@ Secret Key URI `//Alice` is account:
   SS58 Address:     5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
 ```
 
-There is nothing special about the name Alice. You can do the same trick with your own name. But
-remember that keys like these are _not_ secure, and are only useful for experimenting.
+A mnemonic phrase is a secret group of words that represent a wallet. When used in sequence they allow access to the storage within. Specifically there is a dev phrase implemented whenever a mnemonic phrase is omitted.```Alice``` for example is:
```
shawntabrizi, line 325:
```suggestion
A mnemonic phrase is a secret group of words that can be used to uniquely generate a private key. In Subkey, there is a specific dev phrase used whenever a mnemonic phrase is omitted.```//Alice``` for example is actually:
```

PR #617, opened 2020-08-05
Align POE tutorial with template

riusricardo:
Remove the system alias and aligns the tutorial to the node-template code.


JoshOrndorff:
This is a good change that I would like to see merged. Reviewing it led me to realize that this alias is not entirely removed from the template yet. I've begun that process in https://github.com/paritytech/substrate/pull/6829

PR #618, opened 2020-08-05
Minor clarification about finality voting

JoshOrndorff:
This PR correct a minor mistake to the note about finality that was added in #610.

As a practical counter-example, try running this tutorial and fail to insert Bob's Aura key. You'll find that only Alice authors blocks, but both authorities cast finality votes. Blocks will still be finalized.


danforbes:
Thank you, Joshy! I added some colors to the docs by describing the purpose of the Aura key. Please let me know if it's accurate.

JoshOrndorff:
Looks fine to me.

PR #619, opened 2020-08-06
Add Some Diagrams (and Supporting Materials) Created for Workshops

danforbes:
This PR includes a Prettier commit. This is the commit I'm looking for the most feedback on https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/619/commits/0bc84aef047e747a57d955fbd4056ab7b3e33eba


docs/index.md
```diff
@@ -6,69 +6,87 @@ This page is the top level entry point into the Substrate documentation.
...
+  compiled to [Wasm](knowledgebase/getting-started/glossary#webassembly-wasm) and becomes part of the blockchain's
+  storage state - this enables one of the defining features of a Substrate-based blockchain:
+  [forkless runtime upgrades](knowledgebase/advanced/executor#forkless-runtime-upgrades). Substrate clients may also
+  include a "native runtime" that is compiled for the same platform as the client itself (as opposed to Wasm).
+- **Peer-to-peer network** capabilities allow the client to communicate with other network participants. Substrate uses
+  [the `libp2p` network stack](https://libp2p.io/).
+- **Consensus** engines provide logic that allows network participants to agree on the state of the blockchain.
+  Substrate makes it possible to supply custom consensus engines and also ships with several consensus mechanisms that
+  have been built on top of [Web3 Foundation research](https://w3f-research.readthedocs.io/en/latest/index.html).
+- **Remote procedure calls** allow blockchain users to interact with the network. Substrate provides HTTP and WebSocket
```
shawntabrizi, line 43: No one actually uses the term "Remote procedure calls" right? Just say RPC, which seems pretty colloquial.

docs/index.md
```diff
@@ -6,69 +6,87 @@ This page is the top level entry point into the Substrate documentation.
...
+![Substrate Client Architecture](assets/substrate-arch.png)
+
+The Substrate client is an application that runs a Substrate-based blockchain node - it consists of several components
+that include, but are not limited to, the following:
+
+- **Storage** is used to persist the evolving state of the decentralized system represented by a blockchain. The
+  blockchain network allows participants to reach trustless [consensus](knowledgebase/advanced/consensus) about the
+  state of storage. Substrate ships with a simple and highly efficient
+  [key-value storage mechanism](knowledgebase/advanced/storage).
+- **Runtime** logic defines how blocks are processed, including state transition logic. In Substrate, runtime code is
```
shawntabrizi, line 33: The emphasis I would put on the substrate client is that it contains a **Wasm interpreter** which is used to execute the runtime. Either an additional bullet point above this about the wasm interpreter or in replacement of this line makes sense to me.

docs/index.md
```diff
@@ -6,69 +6,87 @@ This page is the top level entry point into the Substrate documentation.
...
+  compiled to [Wasm](knowledgebase/getting-started/glossary#webassembly-wasm) and becomes part of the blockchain's
+  storage state - this enables one of the defining features of a Substrate-based blockchain:
+  [forkless runtime upgrades](knowledgebase/advanced/executor#forkless-runtime-upgrades). Substrate clients may also
```
shawntabrizi, line 36: I dont believe this is the proper way to do Docusarus links

https://docusaurus.io/docs/en/navigation

You should target the `md` file which will then be processed by docusaurus to generate the correct url. Versus here, you are assuming how docusaurus will generate the link, and this will fail for things like versioning or whatever other fringe scenarios the documentation has.

docs/index.md
```diff
@@ -6,69 +6,87 @@ This page is the top level entry point into the Substrate documentation.
...
-   is built and should suffice for almost all needs in the near future. For a tutorial on this, see
+1. **With the Substrate Node**: You can run the pre-designed Substrate Node and configure it with a genesis block that
+   includes the default node runtime. In this case, you just need to configure a JSON file and launch your own
+   blockchain. This affords you the least amount of customization, only allowing you to change the genesis parameters of
+   the included runtime modules such as: balances, staking, block-period, fees, governance, etc... For a tutorial on
+   doing this, see [Start a Private Network with Substrate](tutorials/start-a-private-network/index.md).
+
+2. **With Substrate FRAME**: You can easily create your own custom runtime using FRAME. This affords you a very large
+   amount of freedom over your own blockchain's logic, letting you change data types, select from the library of
+   modules, and add your own custom modules. Much can be changed without touching the block-authoring logic since it is
```
shawntabrizi, line 61: modules should potentially be `pallets` depending on how you are standardizing terminology in the docs.

I dont see much difference between the two personally

docs/index.md
wheresaddie: I might suggest doing both- say RPC (Remote Procedure Calls) just to define it on its first instance for newbies

docs/index.md
wheresaddie: technical is there any deviation in the terms/implementation or are they interchangeable? I ask for future style guides

shawntabrizi: Well there is historical and literal reasons to use the term module.

Before we had "FRAME" and "Pallet", it was "SRML (Substrate Runtime Module Library)" and "Modules"

Crates and Modules are the official terms Rust uses "split code into logical units":
https://doc.rust-lang.org/1.0.0/book/crates-and-modules.html#basic-terminology:-crates-and-modules

So all Pallets are Rust modules/crates, but not all crates/modules are Pallets.

Also a pallet may itself have modules like the benchmarking, test, or mock module which are imported with:

```
mod test;
mod benchmarking;
mod mock;
```

etc...


So its all about specificity and whatever you think is best.

shawntabrizi: Probably best to go with "pallet" here on second review.

docs/index.md
danforbes: Probably true, but I'm going to stick with the existing conventions for now. We can make a separate sweep of internal links if we deem that worthwhile.

docs/index.md
danforbes: This is one of the sections I struggled with the most. You are undoubtedly technically correct, but I'm not sure if what you're describing is the most constructive abstraction to create for our users. For one thing, the interpreter is not included in the diagram and I'm trying to keep this list of components restricted to what appears in the diagram. I also think defining the concept of a runtime and describing that, in Substrate, it is compiled to a Wasm blob and included as part of the chain state, is more important than the executor/interpreter, which I see as more of an implementation/optimization detail.

shawntabrizi: The "substrate client" is the executable that runs a node.

The wasm interpreter included with it is certainly more important than the included runtime. You can pass to your substrate client a chain specification with a runtime that is different than the one included in the client, and it will execute that wasm runtime in the chainspec instead of the one compiled into the client.

For anyone to write a Substrate/Polkadot client in another language would not need to include the runtime information into the client, just provide a Wasm executor.

The runtime being included into the client is the optimization, and optional piece.

danforbes: > The wasm interpreter included with it is certainly more important than the included runtime.

I think this is where we disagree.

shawntabrizi: I am not sure where to go from here. You should ask for feedback in the public channels and get other people's opinion then.

Take a look at the `client` folder in Substrate: https://github.com/paritytech/substrate/tree/master/client

danforbes: Let's see what the other members of the Developer Hub team think

JoshOrndorff:
```suggestion
- **Runtime Executor**  executed the runtime logic which defines how blocks are processed, including state transition logic. In Substrate, runtime code is
```

Here's an attempt at a compromise. I think Shawn's point is valid ad the executor is worth mentioning. I also think Dan's point about it being too fine of a detail to fit what he's explaining here is valid. Ultimately I think this PR is a big improvement, and I'm fine with it going in as is, with my suggestion, or with some other rewording.

docs/index.md
JoshOrndorff: I agree that pallet is better here.

docs/index.md
danforbes: I added two sentences to the end of this bullet point that I hope covers all our bases in a way that is complete _and_ easy to understand https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/619/commits/9ad51ffe215f4fd8c03c77b6c0e371f6b548f914 It's a little on the verbose side but I think it has the nice "side effect" of adding color to the forkless runtime comment.

docs/index.md
```diff
@@ -6,69 +6,91 @@ This page is the top level entry point into the Substrate documentation.
...
+  [key-value storage mechanism](knowledgebase/advanced/storage).
+- **Runtime** logic defines how blocks are processed, including state transition logic. In Substrate, runtime code is
+  compiled to [Wasm](knowledgebase/getting-started/glossary#webassembly-wasm) and becomes part of the blockchain's
+  storage state - this enables one of the defining features of a Substrate-based blockchain:
+  [forkless runtime upgrades](knowledgebase/advanced/executor#forkless-runtime-upgrades). Substrate clients may also
+  include a "native runtime" that is compiled for the same platform as the client itself (as opposed to Wasm). The
+  component of the client that dispatches calls to the runtime is known as the
+  [executor](knowledgebase/advanced/executor) and it selects between the native code and interpreted Wasm. Although the
+  native runtime may offer a performance advantage, the executor will select to interpret the Wasm runtime if it
+  implements a newer [version](knowledgebase/advanced/executor#runtime-versioning).
```
jimmychu0807, line 41: I think this paragraph pretty much resolve the difference of pov between Dan and Shawn above.

docs/index.md
jimmychu0807:
```suggestion
2. **With Substrate FRAME**: You can easily create your own custom runtime using Framework for Runtime Aggregation of Modularized Entities (FRAME). This affords you a very large
   amount of freedom over your own blockchain's logic, letting you change data types, select from the library of
   pallets, or even add your own pallets. Much can be changed without touching the block-authoring logic since it is
```

jimmychu0807: just to make the task of replacement easier 😄

docs/index.md
```diff
@@ -6,69 +6,91 @@ This page is the top level entry point into the Substrate documentation.
...
-   syncing. If the block authoring logic needs to be modified, then a new block-authoring binary
-   must be built as a separate project and used by validators. This is how the Polkadot relay chain
-   is built and should suffice for almost all needs in the near future. For a tutorial on this, see
+1. **With the Substrate Node**: You can run the pre-designed Substrate Node and configure it with a genesis block that
+   includes the default node runtime. In this case, you just need to configure a JSON file and launch your own
+   blockchain. This affords you the least amount of customization, only allowing you to change the genesis parameters of
+   the included runtime modules such as: balances, staking, block-period, fees, governance, etc... For a tutorial on
+   doing this, see [Start a Private Network with Substrate](tutorials/start-a-private-network/index.md).
+
+2. **With Substrate FRAME**: You can easily create your own custom runtime using FRAME. This affords you a very large
```
jimmychu0807, line 63:
```suggestion
2. **With Substrate FRAME**: You can easily create your own custom runtime using the Framework for Runtime Aggregation of Modularized Entities (FRAME). This affords you a very large
```

jimmychu0807: we probably have never defined what the term **FRAME** stands for in this article. I know, it is mouthful.

docs/index.md
```diff
@@ -54,28 +54,36 @@ that include, but are not limited to, the following:
...
+
+2. **With Substrate FRAME**: You can easily create your own custom runtime using [FRAME](knowledgebase/runtime/frame.md)
+   (Framework for Runtime Aggregation of Modularized Entities), which is the method used by the Substrate Node. This
+   affords you a very large amount of freedom over your blockchain's logic, and allows you to configure data types,
+   select from a library of modules (called "pallets"), and add your own custom pallets. Much can be changed without
+   touching the block-authoring logic since it is directed through runtime logic. If this is sufficient, then the
+   existing Substrate binary can be used for block authoring and syncing; the
+   [Substrate Developer Hub Node Template](https://github.com/substrate-developer-hub/substrate-node-template) is a
+   helpful starting point for projects with such needs. If the block authoring logic needs to be modified, then a new
+   block-authoring binary must be built as a separate project and used by validators. This is how the
```
JoshOrndorff, line 75: This bit about block authoring seems too in-the-weeds to me. Basically this is a list of three ways to get the runtime you want. Dipping into block authoring is kind of an orthogonal thing. And even if you do want to keep it it isn't correct that the authoring would happen in a separate binary.

JoshOrndorff: Oh I see. You mean the binary with modified authoring logic would be separate from the node template? That's true.

PR #621, opened 2020-08-10
Correct and clarify docs about `subkey insert`

JoshOrndorff:
This PR adds details and examples about using `subkey insert`. Specifically it clarifies that the `-e` option is necessary when inserting grandpa ed25519 keys. Reported by @siman


siman:
Looks good.

PR #624, opened 2020-08-12
Fix FRAME runtime diagram typo and replace old diagram

danforbes:


danforbes:
@JoshOrndorff - I updated the new FRAME runtime diagram with a newer one that corrected a typo; there was one instance of `Traesury`.

PR #656, opened 2020-09-08
Use consistent naming for Subkey, etc

danforbes:


JoshOrndorff (approved):
This is an improvement, but this entire integrate/subkey article is written like a tutorial, not a kb article.

Issue #655, opened 2020-09-07
Remove Joshy's face from Seminar Page


JoshOrndorff:
Over the past few months, the seminar has transitioned from being my thing to more of a DevHub Team thing. I think this is a good direction for the seminar to continue heading in. To that end, my face should be removed from the Seminar page.

wheresaddie:
#666 ftw lol

PR #658, opened 2020-09-08
Update Seminar Page

JoshOrndorff:
This PR removes the September 8th Seminar topic from the seminar page. It does not schedule any new topics, but this is not a problem because seminar topics are still scheduled three weeks out.

### DIFF
--- a/website/pages/en/seminar.js
+++ b/website/pages/en/seminar.js
@@ -104,9 +104,6 @@ function Seminar(props) {
 						<div className='col-md-6 first'>
 							<h4 className='mb-2'>What topics are coming up?</h4>
 							<p>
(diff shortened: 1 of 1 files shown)


wheresaddie (approved):
also for host lead Id just suggest we use DevHub Team vs any specific persons name from now on.. if that's ok with everyone--

danforbes:
> also for host lead Id just suggest we use DevHub Team vs any specific persons name from now on.. if that's ok with everyone--

Strongly agree :+1:

JoshOrndorff:
> also for host lead Id just suggest we use DevHub Team vs any specific persons name from now on.. if that's ok with everyone--

Do you want to make a PR that updates the page in this way? It would solve #655

wheresaddie:
going out with a bang @JoshOrndorff #666

Issue #580, opened 2020-06-20
Better Runtime Upgrade Documentation


danforbes:
We can create more [documentation around runtime upgrades](https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/blob/source/docs/knowledgebase/advanced/executor.md#runtime-upgrades), which is one of Substrate's most important and distinguishing characteristics. In particular, a community member recently asked for an enumeration of the changes that _can_ be included in a  forkless runtime upgrade and those that _cannot_.

wheresaddie:
yes.. great suggestion.. ! will put this on the list for next week's DevHub to-dos. thanks for the suggestion @danforbes

JoshOrndorff:
Some existing (but possibly slightly outdated) resources you may find helpful:
https://substrate.dev/substrate-beginner-workshop/#/
https://www.youtube.com/watch?v=0eNGZpNkJk4&list=PLp0_ueXY_enXRfoaW7sTudeQH10yDvFOS (also features storage migrations)
https://www.youtube.com/watch?v=yHodYke7nms&list=PLp0_ueXY_enXRfoaW7sTudeQH10yDvFOS

danforbes:
Comments from @HCastano on how Knowledge Base docs can complement #670 

    How does the node know to use the version on chain instead of the native version?
    Why does the runtime + client split allow for this? (Might be helpful to go into the responsibilities of each)
    Under what circumstances would I need to restart my node?
    Potential "gotcha's", like not updating the spec_version

danforbes:
The enhanced documentation should include the materials created by @apopiak https://hackmd.io/BQt-gvEdT66Kbw0j5ySlWw

wheresaddie:
thanks for tackling this @danforbes I made some minor PRs on #750 and appreciate the additional context and documentation <3

apopiak:
@danforbes Do you want to use a new issue to track things like a runtime migration tutorial or the above mentioned "things that can and cannot be changed via runtime upgrade" or do we want to reopen this one?

wheresaddie:
Id start new ones @apopiak @danforbes otherwise its just a link to a link to a link until we all die

PR #670, opened 2020-09-17
Upgrade a Chain tutorial

danforbes:


docs/tutorials/upgrade-a-chain/index.md
```diff
@@ -0,0 +1,47 @@
+---
+title: Introduction
+---
+
+One of the defining features of the Substrate blockchain development framework is its support for
+forkless runtime upgrades. This means that Substrate-based networks can be enhanced in a way that is
+supported and protected by the capabilities of the blockchain itself. A blockchain's
+[runtime](../../knowledgebase/runtime) is the mechanism by which
+[state](../../knowledgebase/runtime/storage) changes are effected. Because the definition of the
```
joepetrowski, line 9: Two gross sentences. "This ____ means...", "enhanced in a way that is", "changes are effected."

docs/tutorials/upgrade-a-chain/index.md
```diff
@@ -0,0 +1,47 @@
...
+[existential (minimum) balance](../../knowledgebase/getting-started/glossary#existential-deposit)
+for network accounts.
+
+If you have problems with this tutorial, the Substrate community is full of helpful resources. We
+maintain an active
+[#SubstrateTechnical chat room](https://app.element.io/#/room/!HzySYSaIhtyWrwiwEV:matrix.org) and
+monitor the
+[`substrate` tag on Stack Overflow](https://stackoverflow.com/questions/tagged/substrate). You can
+also use the [`subport` GitHub repository](https://github.com/paritytech/subport/issues/new) to
+create an issue. We encourage community to be actively part of the future of Substrate.
```
joepetrowski, line 35: Active verbs when possible
```suggestion
create an issue. We encourage the community to actively participate in the future of Substrate. 
```

docs/tutorials/upgrade-a-chain/sudo-upgrade.md
```diff
@@ -0,0 +1,163 @@
...
+	authoring_version: 1,
+	spec_version: 2,  // Update this value.
+	impl_version: 1,
+	apis: RUNTIME_API_VERSIONS,
+	transaction_version: 1,
+};
+```
+
+Build the upgraded runtime.
+
```
jimmychu0807, line 140: Maybe briefly explain about what `authoring_version`, `spec_version`, `impl_version`, and `transaction_version` here as well. When to upgrade which value, or reference to places where readers can find the answer.

docs/tutorials/upgrade-a-chain/sudo-upgrade.md
```diff
@@ -0,0 +1,163 @@
...
+```shell
+cargo build --release -p node-template-runtime
+```
+
+## Upgrade the Runtime
+
+Use this link to open the Polkadot JS Apps UI and automatically configure the UI to connect to the
+local node: https://polkadot.js.org/apps/#/extrinsics?rpc=ws://127.0.0.1:9944. Use Alice's account
+to invoke the `sudo_unchecked_weight` function and use the `set_code` function from the System
+pallet as its parameter. In order to supply the build artifact that was generated by the previous
```
jimmychu0807, line 150:
```suggestion
pallet as its parameter. The reason we can use Alice account to make the sudo call is because we have added Alice private key in Sudo keystore when we run the node in development mode (started with `--dev` flag). In order to supply the build artifact that was generated by the previous
```

docs/tutorials/upgrade-a-chain/sudo-upgrade.md
```diff
@@ -0,0 +1,163 @@
...
+function's weight annotation also specifies that `set_code` is in
+[the `Operational` class](../../knowledgebase/runtime/fees#operational-dispatches) of dispatchable
+functions, which identifies it as relating to network _operations_ and impacts the accounting of its
+resources, such as by exempting it from the
+[`TransactionByteFee`](https://substrate.dev/rustdocs/v2.0.0-rc6/pallet_transaction_payment/trait.Trait.html#associatedtype.TransactionByteFee).
+In order to work within FRAME's safeguards around resource accounting, the Sudo pallet provides the
+[`sudo_unchecked_weight`](https://substrate.dev/rustdocs/v2.0.0-rc6/pallet_sudo/enum.Call.html#variant.sudo_unchecked_weight)
+function, which provides the same capability as the `sudo` function, but accepts an additional
+parameter that is used to specify the (possibly zero) weight to use for the call that is dispatched
+from the `Root` origin. The `sudo_unchecked_weight` function is what will be used to invoke the
```
joepetrowski, line 61: Weight doesn't get passed into downstream calls, only the top-level dispatchable. In this case, the weight of `sudo_unchecked_weight` and whatever is wrapped inside of it get passed as one to the block constructor.
Namely, no weight going in here:
`let res = call.dispatch_bypass_filter(frame_system::RawOrigin::Root.into());`

docs/tutorials/upgrade-a-chain/scheduled-upgrade.md
```diff
@@ -0,0 +1,79 @@
...
+	type DustRemoval = ();
+	type ExistentialDeposit = ExistentialDeposit;
+	type AccountStore = System;
+	type WeightInfo = ();
+}
+```
+
+This change increases the value of the Balances pallet's
+[`ExistentialDeposit`](../../knowledgebase/getting-started/glossary#existential-deposit) - the
+minimum balance needed to keep an account alive from the point-of-view of the Balance's pallet. Keep
```
joepetrowski, line 48:
```suggestion
minimum balance needed to keep an account alive from the point-of-view of the Balances pallet. Keep
```

jimmychu0807:
Dan, I think the tutorial looks good. Thanks for the great work. 👍

jimmychu0807 (approved):

docs/tutorials/upgrade-a-chain/scheduled-upgrade.md
```diff
@@ -0,0 +1,79 @@
...
+In the previous section, the Scheduler pallet was configured with the `root` origin as its
+[`ScheduleOrigin`](https://substrate.dev/rustdocs/v2.0.0-rc5/pallet_scheduler/trait.Trait.html#associatedtype.ScheduleOrigin),
+which means that the `sudo` function (_not_ `sudo_unchecked_weight`) can be used to invoke the
+`schedule` function. Use this link to open the Polkadot JS Apps UI's Sudo tab:
+https://polkadot.js.org/apps/#/sudo?rpc=ws://127.0.0.1:9944. Wait until all the other fields have
+been filled in before providing the `when` parameter. Leave the `maybe_periodic` parameter empty and
+the `priority` parameter at its default value of `0`. Select the System pallet's `set_code` function
+as the `call` parameter and provide the Wasm binary as before. Leave the "with weight override"
+option deactivated. Once all the other fields have been filled in, use a block number about 5 blocks
+(30 seconds) in the future to fill in the `when` parameter and quickly submit the transaction. You
```
joepetrowski, line 69: For someone going through step by step, 5 blocks is really cutting it close.

jimmychu0807:
See if you need to address the link-check failure in the CI.

docs/tutorials/upgrade-a-chain/scheduled-upgrade.md
```diff
@@ -0,0 +1,79 @@
...
+can use the template node's command line output or the
+[Polkadot JS Apps UI block explorer](https://polkadot.js.org/apps/#/explorer?rpc=ws://127.0.0.1:9944)
+to select a block number.
+
+![Scheduled Upgrade](assets/tutorials/upgrade-a-chain/scheduled-upgrade.png)
+
+After the target block has been included in the chain, the version number in the upper-left-hand
+corner of Polkadot JS Apps UI should reflect that the runtime version is now `3`.
```
joepetrowski, line 77: I would add a call to action to show that the chain's logic actually changed. Like, "try sending a balance transfer of 700, which totally would have succeeded before, but now will fail." It's more exciting than "the spec number changed _on the UI_."

docs/tutorials/upgrade-a-chain/index.md
danforbes: I did my best to address this comment, but next time something a little more constructive/specific would be appreciated :pray:

docs/tutorials/upgrade-a-chain/sudo-upgrade.md
danforbes: Do you think that anything in this tutorial needs to be updated to clarify this?

joepetrowski: Well this is not clear:

> accepts an additional ￼parameter that is used to specify the (possibly zero) weight to use for the _call that is dispatched ￼from the `Root` origin._

The call that is dispatched from the root origin doesn't have any weight. I.e., the `_weight` param is for the weight of the `sudo_unchecked_weight` function, including the `call`, but not the weight for the `call`.

docs/tutorials/upgrade-a-chain/index.md
joepetrowski: It's better, I would still try to be more active. "A runtime defines the information that your blockchain can hold and the logic for effecting changes to that information. Substrate supports the capability to make breaking changes to your runtime without a hard fork." 

Then you say _how_ it supports that capability: "Because the definition of the runtime is itself an element in a Substrate chain's state..."

wheresaddie (approved):

docs/tutorials/upgrade-a-chain/sudo-upgrade.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+## Start the Template Node
+
+Since forkless runtime upgrades do not require network participants to restart their blockchain
+clients, the first step of this tutorial is to start the template node as-is. Build and start the
+unmodified [Node Template](https://github.com/substrate-developer-hub/substrate-node-template).
+
+```shell
+cargo run --release -- --dev --tmp
```
HCastano, line 38: For a tutorial do we really need users to run their node in `release` mode? It's just going to add extra time to their compilation. I know that `set_code` can take a long time, but I imagine that for a local chain it'll be fine with no optimizations

docs/tutorials/upgrade-a-chain/sudo-upgrade.md
```diff
@@ -0,0 +1,182 @@
...
+which is a member of
+[the `RuntimeVersion` struct](https://substrate.dev/rustdocs/v2.0.0-rc6/sp_version/struct.RuntimeVersion.html)
+that is defined in `runtime/src/lib.rs`.
+
+```rust
+pub const VERSION: RuntimeVersion = RuntimeVersion {
+	spec_name: create_runtime_str!("node-template"),
+	impl_name: create_runtime_str!("node-template"),
+	authoring_version: 1,
+	spec_version: 2,  // Update this value.
```
HCastano, line 139: I think it'll be worthwhile to take a second and explain why this is required, and what happens if you don't update it

docs/tutorials/upgrade-a-chain/sudo-upgrade.md
```diff
@@ -0,0 +1,182 @@
...
+  // Dispatch the specified call as the Root origin
+  let res = call.dispatch(Origin::Root);
+  Ok()
+}
+```
+
+## Start the Template Node
+
+Since forkless runtime upgrades do not require network participants to restart their blockchain
+clients, the first step of this tutorial is to start the template node as-is. Build and start the
```
HCastano, line 34: I would also a a note here saying that they should leave this node running throughout the tutorial since this is what shows the forkless nature of forkless upgrades

HCastano (commented):
I feel like we're missing some of the fundamentals of how forkless upgrades work. 

- How does the node know to use the version on chain instead of the native version?
- Why does the runtime + client split allow for this? (Might be helpful to go into the responsibilities of each)
- Under what circumstances would I need to restart my node?
- Potential "gotcha's", like not updating the `spec_version` (which ties into my first point)

I'm glad to see a tutorial on this is being written though, I'm surprised it wasn't already a thing!

joepetrowski:
> I feel like we're missing some of the fundamentals of how forkless upgrades work.
> 
> * How does the node know to use the version on chain instead of the native version?
> * Why does the runtime + client split allow for this? (Might be helpful to go into the responsibilities of each)
> * Under what circumstances would I need to restart my node?
> * Potential "gotcha's", like not updating the `spec_version` (which ties into my first point)
> 
> I'm glad to see a tutorial on this is being written though, I'm surprised it wasn't already a thing!

This will turn into an essay and I think that tutorials should be more active and the user should feel that they are regularly making progress. This is the kind of documentation that should be in the knowledge base.

docs/tutorials/upgrade-a-chain/sudo-upgrade.md
danforbes: This is a really great comment, @HCastano :pray: I kind of always just took for granted that all our docs specified a release build. I just did [a little research](https://users.rust-lang.org/t/why-does-cargo-build-not-optimise-by-default/4150/2) and it completely makes sense that we'd want to omit this flag in tutorials and such.

danforbes:
Noting that this PR relates to #580

docs/tutorials/upgrade-a-chain/index.md
danforbes: "Active voice" is a writing concept that ISWTG I _cannot_ wrap my head around :sweat_smile:

docs/tutorials/upgrade-a-chain/sudo-upgrade.md
danforbes: I'm concerned that you're actually sacrificing clarity for strict technical correctness. Can you suggest a rewording that is as clear and more technically correct?

docs/tutorials/upgrade-a-chain/sudo-upgrade.md
joepetrowski: I thought that compiling without optimizations will result in the binary not even able to produce blocks? It's really only good for checking if it does compile.

docs/tutorials/upgrade-a-chain/sudo-upgrade.md
joepetrowski: Hmm, I got hung up on this sentence because it felt unclear to which call the `_weight` param applied to. I would just end the sentence earlier:

"...but accepts an additional parameter that is used to specify the (possibly zero) weight to use for the call."

docs/tutorials/upgrade-a-chain/sudo-upgrade.md
danforbes: I definitely checked :+1: I can confirm that the node produces and finalizes blocks :heavy_check_mark:

HCastano: > It's really only good for checking if it does compile.

@joepetrowski There's a special cargo command for this, [`cargo check`](https://doc.rust-lang.org/cargo/commands/cargo-check.html), which doesn't build a binary but just runs through part of the compilation process to make sure your code compiles

PR #674, opened 2020-09-19
Integration Docs on Transaction Encoding

danforbes:


docs/knowledgebase/integrate/encode-txs.md
```diff
@@ -0,0 +1,58 @@
...
+
+- Signed (additional) information: data that should be well-known to both the signer and verifier,
+  such as the hash of the chain's genesis block; does not contribute to payload size because it is
+  signed (hashed)
+- Unsigned (extra) information: data that only the signer knows and that must be sent in the clear,
+  such as the nonce used for a transaction; will contribute to the size of a payload
+
+## Signing
+
+- 1 byte: magic number that indicates signing scheme; `00` for Ed25519, `ff` for sr25519
```
joepetrowski, line 48: ECDSA?

docs/knowledgebase/integrate/encode-txs.md
```diff
@@ -0,0 +1,58 @@
...
+- 64 bytes: sig(call + unsigned/extra + signed/additional)
+
+## Transaction Construction
+
+Given the above, transactions are encoded as follows:
+
+- variable: SCALE compact integer-encoded length of the entire transaction
+- variable: dispatchable call
+- 65 bytes: signed data
+- variable: signed extension unsigned (extra) information
```
joepetrowski, line 58: - 4 bytes: spec version
- 4 bytes: transaction version
- 32 bytes: genesis hash
- 32 bytes: block hash

docs/knowledgebase/integrate/encode-txs.md
```diff
@@ -0,0 +1,58 @@
+---
+title: Encoding Transactions
```
joepetrowski, line 2:
```suggestion
title: Transaction Serialization
```

docs/knowledgebase/integrate/encode-txs.md
```diff
@@ -0,0 +1,58 @@
...
+- 1 byte: 0-based call index that corresponds to the target variant on the module's `Call` enum
+- variable: [SCALE](../../knowledgebase/advanced/codec)-encoded function parameters
+- 32 bytes: [address](../../knowledgebase/learn-substrate/account-abstractions) of function caller
+
+## Signed Extensions
+
+Substrate uses
+"[signed extensions](../../knowledgebase/learn-substrate/extrinsics#signed-extension)" to allow
+developers to extend its transaction-verification capabilities. Despite the name, signed extensions
+can also be applied to unsiqned transactions. They are intended to be used in any situation where it
```
joepetrowski, line 27:
```suggestion
can also be applied to unsigned transactions. They are intended to be used in any situation where it
```

wheresaddie (approved):
once Joe's mods are committed I see no further issues

joepetrowski:
> once Joe's mods are committed I see no further issues

Sorry, but this needs a lot more work to be complete. Please hold off on merging.

docs/knowledgebase/integrate/encode-txs.md
danforbes: I got the terminology here https://wiki.polkadot.network/docs/en/learn-cryptography
> Sr25519 is based on the same underlying Curve25519 as its EdDSA counterpart, Ed25519.

docs/knowledgebase/integrate/encode-txs.md
danforbes: This is dependent upon the signed extensions, though, right? This isn't part of the spec, per se. Or am I wrong?

docs/knowledgebase/integrate/encode-txs.md
danforbes: Why?

docs/knowledgebase/integrate/encode-txs.md
danforbes: Stop trying to extinquish my creativity, Joe :expressionless:

docs/knowledgebase/integrate/encode-txs.md
joepetrowski: Sure, but Substrate also supports ECDSA. You need ed25519, sr25519, **and** ECDSA.

docs/knowledgebase/integrate/encode-txs.md
joepetrowski: Ah yeah you might be right. Well we need to document how to figure out what "variable" information is there.

docs/knowledgebase/integrate/encode-txs.md
danforbes: Ah, I see - do you know what the magic number is for that scheme?

joepetrowski: No, I do not :stuck_out_tongue_closed_eyes:

danforbes:
Ref: https://github.com/darkfriend77/SubstrateNetApi/blob/d6349a46ec23568510d8b238562391e150ca8095/SubstrateNetApi/MetaDataModel/Extrinsic/ExtrinsicModel.cs#L31

stiiifff:
Don't know if that helps, while troubleshooting an issue on the enterprise sample, I went through the process of deconstructing a signed tx. Maybe you could use a similar example e.g. a balance transfer

```
Example signed extrinsic
0xE101844A06DB21B46DFF341B234D8E421C215F32C7799481E5D39DD193D00B5ABD976C01903142CD7BC28A94733541CBE2E5C8553253B820ACFA06AFB7F5E8568F70C82EFEF2C5D00061D477EF5E8DB43FE54C04AA91B2ADDA2288882A1DB82F5EC41183330000000A01145330303032000B10475BD474010000

Extr. Length: E101 (scale-encoded) => 120 bytes
SignVersion : 84 (4 Tx version | 0x80 for signed tx)
PubKey      : 4A06DB21B46DFF341B234D8E421C215F32C7799481E5D39DD193D00B5ABD976C
Signature
    Type    : 01 (Sr25519)
    Sign    : 903142CD7BC28A94733541CBE2E5C8553253B820ACFA06AFB7F5E8568F70C82EFEF2C5D00061D477EF5E8DB43FE54C04AA91B2ADDA2288882A1DB82F5EC41183
Extra
    Era     : 3300
    Nonce   : 00
    Fee     : 00
Call
    Module  : 0A (product_tracking)
    Method  : 01 (track_shipment)
    Params  : 145330303032000B10475BD474010000
```

docs/knowledgebase/integrate/encode-txs.md
```diff
@@ -0,0 +1,58 @@
...
+The purpose of this document is to describe the specification for encoding
+[transactions](../../knowledgebase/learn-substrate/extrinsics) to submit to Substrate-based chains.
+
+## Dispatchable Calls
+
+Transactions on Substrate-based chains are used to invoke
+"[dispatchable calls](../../knowledgebase/getting-started/glossary#dispatch)", which are encoded as
+follows:
+
+- 1 byte: `04`, magic number that identifies a dispatchable call
```
bkchr, line 14: Not sure where you have that from?

danforbes: It's what happens when you try to reverse engineer an opaque encoding :wink:

bkchr: Why opaque? The encoding is written in our Rust code?

danforbes: I am not sure what the purpose of this discussion is.

danforbes:
I've pushed some updates. @stiiifff do you have more details on the signature version? Where can I find that in the Rustdocs?

shawntabrizi:
Some text copied from when i posted in Element Chat:

---

roxontox: I have pushed another update which has even more stuff in it you need. I was wondering where all those 0 at the end come from, but this is the "additional signed" data.

So the output is now:
```
Extra: (CheckSpecVersion, CheckTxVersion, CheckGenesis, CheckMortality(Era::Mortal(2048, 99)), CheckNonce(0), CheckWeight, ChargeTransactionPayment<0>)
Extra Encoded: [58, 6, 0, 0]
Additional Signed: Ok((259, 1, 0x0000000000000000000000000000000000000000000000000000000000000000, 0x0000000000000000000000000000000000000000000000000000000000000000, (), (), ()))
Additional Signed Encoded: [0, 3, 1, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
Raw Payload: [6, 0, 255, 212, 53, 147, 199, 21, 253, 211, 28, 97, 20, 26, 189, 4, 169, 159, 214, 130, 44, 133, 88, 133, 76, 205, 227, 154, 86, 132, 231, 165, 109, 162, 125, 145, 1, 58, 6, 0, 0, 3, 1, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```

The final payload you want to create is quite simple: `(call, extra, additional_signed)`

so we already know how to get the call bytes, you seem to have done this correctly and we can confirm easily with Polkadot JS.

The extra is now being shown here in this output, just those 4 bytes

the "additional_signed" is probably your missing piece, which is all those items we were ignoring before, and should not actually be included in the payload. Here we see they are:
(spec_version, tx_version, gensis_hash, mortality_hash, nonce, weight, payment), just like we would expect

but for nonce, weight and payment, the `additional_signed` is empty `()` because we have that in the direct "extra encoded"

You can see spec version and runtime version is coming from the runtime:
```
/// Runtime version.
pub const VERSION: RuntimeVersion = RuntimeVersion {
	/* -- snip -- */
	spec_version: 259,
	impl_version: 0,
	apis: RUNTIME_API_VERSIONS,
	transaction_version: 1,
};
```

and additional signed info is coming from the configuration of that signed extension. Here is genesis where we expect a hash as an "additional signed" payload:
```
impl<T: Trait + Send + Sync> SignedExtension for CheckGenesis<T> {
	type AccountId = T::AccountId;
	type Call = <T as Trait>::Call;
	type AdditionalSigned = T::Hash;   <--- Here
	type Pre = ();
	const IDENTIFIER: &'static str = "CheckGenesis";
	fn additional_signed(&self) -> Result<Self::AdditionalSigned, TransactionValidityError> {
		Ok(<Module<T>>::block_hash(T::BlockNumber::zero()))
	}
}
```

And nonce we where we DO expect some extra payload:
```
pub struct CheckNonce<T: Trait>(#[codec(compact)] T::Index);
```

A compact index (u32)

But we dont expect an `AdditionalSigned` payload:
```
impl<T: Trait> SignedExtension for CheckNonce<T> where
	T::Call: Dispatchable<Info=DispatchInfo>
{
	type AccountId = T::AccountId;
	type Call = T::Call;
	type AdditionalSigned = ();    <--- see here
	type Pre = ();
	const IDENTIFIER: &'static str = "CheckNonce";
```

I assume none of this is exposed in the metadata, so you just have to know how to construct this for the blockchain you are interacting with

And then for the final tx, we drop this additional signed information in order to keep the payload smaller

This is the final line of the `create_transaction` function we have been hacking into.
```
let (call, extra, _) = raw_payload.deconstruct();
Some((call, (address, signature.into(), extra)))
```

So the final payload should be:

call + address + signature_of(call + extra + additional_signed) + extra

and this should be accepted by a node and processed correctly

NukeManDan:
Hey guys, This PR is stale, but I think it would be a valuable addition to the devhub. @sacha-l @jimmychu0807 what do you think? Is this worth resurrecting or using as a basis for a new PR based on comments here?

sacha-l:
> Hey guys, This PR is stale, but I think it would be a valuable addition to the devhub. @sacha-l @jimmychu0807 what do you think? Is this worth resurrecting or using as a basis for a new PR based on comments here?

Yes completely, just needs a little love and it'll be good for another review.

NukeManDan:
https://github.com/substrate-developer-hub/substrate-docs/issues/187

PR #681, opened 2020-09-24
Update most tutorials to v2.0.0

danforbes:
Smart contract tutorials not yet updated.

### DIFF
--- a/docs/tutorials/upgrade-a-chain/scheduled-upgrade.md
+++ b/docs/tutorials/upgrade-a-chain/scheduled-upgrade.md
@@ -29,9 +29,11 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 
 parameter_types! {
 	pub const ExistentialDeposit: u128 = 1000;  // Update this value.
+	pub const MaxLocks: u32 = 50;
 }
 
 impl pallet_balances::Trait for Runtime {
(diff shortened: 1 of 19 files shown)


wheresaddie:
ty dan

wheresaddie (approved):

docs/tutorials/upgrade-a-chain/scheduled-upgrade.md
```diff
@@ -52,7 +52,7 @@ dropped - that would require a storage migration, which is out of the scope of t
 Build the upgraded runtime.
 
 ```shell
-cargo build --release -p node-template-runtime
+cargo build -p node-template-runtime
```
jimmychu0807, line 55: @joepetrowski want to get your confirmation. When we do a runtime upgrade, we want to compile the runtime in `release` mode, because the file size of the wasm in debugged mode is a lot bigger than the wasm in released mode. Is this true? I have this impression.

jimmychu0807 (approved):
lgtm. Thanks for the effort, Dan!

I think this PR also show you gain expertise in running cargo cmds.

There is only one minor concern about whether to compile the runtime in `release` or `debug` mode in runtime upgrade, but it shouldn't hinder this PR to be merged.

PR #707, opened 2020-10-05
Update metadata.md

wheresaddie:
updated to 0x0c

### DIFF
--- a/docs/knowledgebase/runtime/metadata.md
+++ b/docs/knowledgebase/runtime/metadata.md
@@ -107,9 +107,9 @@ string before taking a more detailed look at the metadata's decoded format.
 ### Encoded Metadata Format
 
 The hex blob that is returned by the JSON-RPCs `state_getMetadata` method starts with a hard-coded
-magic number, `0x6d657461`, which represents "meta" in plain text. The next piece of data (`0x0b` in
-the example above) represents the metadata version; decoding the hexadecimal value `0x0b` yields the
-decimal value 11, which is
+magic number, `0x6d657461`, which represents "meta" in plain text. The next piece of data (`0x0c` in
+the example above) represents the metadata version; decoding the hexadecimal value `0x0c` yields the
+decimal value 12, which is
 [the version of the Substrate metadata format](https://substrate.dev/rustdocs/v2.0.0/frame_metadata/enum.RuntimeMetadata.html)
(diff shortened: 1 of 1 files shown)


danforbes:
Apparently this is the actual change to the transaction encoding (?) that caused the version bump https://github.com/paritytech/substrate/pull/6969 I'm not sure if this change has been incorporated into our docs.

wheresaddie:
good to know. In that case might run this by @folsen or @gnunicorn ..?

gnunicorn:
@thiolliere should be able to say.

thiolliere:
below in the doc is explained each field of module metadata: name, storage, error, ... but not index.

An explanation is that index is:

index comes from how pallet is defined in construct_runtime. For example:
```
construct_runtime!(
...
{
Pallet: my_pallet::{...}, // index is implicitly 0
MySecondPallet: my_second_pallet::{..} = 2, // index is explicitly 2
MyOtherPallet: my_other_pallet::{..}, // index is implicitly 3 (same rules as rust fieldless enum 
}
)
```

I'm not sure it is interesting to go into such details, just saying it is the index assigned in construct_runtime seems enough.

About the change to the transaction encoding, this document doesn't explain it here, but maybe another document covers it ?

The encoding change is this: before:
```
construct_runtime!(
...
{
Pallet: my_pallet::{Event}, // the variant for Pallet event in the runtime event is at index 0
MySecondPallet: my_second_pallet::{Call}, // the variant for MySecondPallet call in the runtime call is at index 0
MyOtherPallet: my_other_pallet::{Call, Event}, // the variant for MyOtherPallet call in the runtime call is at index 1 and event at index 1
}
)
```
and now
```
construct_runtime!(
...
{
Pallet: my_pallet::{Event}, // the variant for Pallet event in the runtime event is at index 0
MySecondPallet: my_second_pallet::{Call}, // the variant for MySecondPallet call in the runtime call is at index 1
MyOtherPallet: my_other_pallet::{Call, Event}, // the variant for MyOtherPallet call in the runtime call is at index 2 and event at index 2
}
)
```
i.e. for Event, Call and Origin the index for the corresponding variant is the index of the pallet.

danforbes:
Is this still relevant or was it superseded by https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/718?

PR #722, opened 2020-10-15
Hopefully fix the javascript that missed a punctuation in a codeblock

jimmychu0807:
Hopefully fix substrate-developer-hub/substrate-front-end-template#138

The reason saying hopefully is because this is not the problem in the javascript in the markdown file, but we used to have a <div></div> wrapping a markdown code block. The whole content is then sent to crowdin for translation and sent back. So the not-so-legal syntax (having markdown code block inside html tag) got misinterpretted along the way and missed the punctuation.


wheresaddie (approved):
ty @jimmychu0807

PR #723, opened 2020-10-15
Replace Gossamer with specific Go implementation

Swader:
Gossamer is a host implementation, hard to use just SCALE from it. A standalone Go SCALE codec is a better fit imo.


riusricardo (approved):
:+1:

PR #727, opened 2020-10-17
Use `master` branch of template repos

danforbes:


riusricardo (changes requested):
I'll be against this until we get proper versioning on the DevHub. :hourglass_flowing_sand:

danforbes:
@riusricardo can you elaborate on and justify those statements?

danforbes:
This is related to @NukeManDan's comment here https://github.com/substrate-developer-hub/substrate-node-template/issues/90#issuecomment-710617694

riusricardo:
> This is related to @NukeManDan's comment here [substrate-developer-hub/substrate-node-template#90 (comment)](https://github.com/substrate-developer-hub/substrate-node-template/issues/90#issuecomment-710617694)

I understood the intention of this PR since the beginning. However this is not the way to deal with the issue you are referencing.
This will be a huge regression. We started this way and found tons of issues. Because of this, we created the `pre-v2.0` version. 
This will become a bigger problem at the time we start versioning again docs and tutorials. Imagine that version `v2.0` will say use `master`, then version `v2.5` will say use `master`. Users will get issues referencing `master` at multiple commits and it will be much more difficult for us to help them.

The DevHUb node-template should follow the Substrate repo as close as possible. For example, this [#81](https://github.com/substrate-developer-hub/substrate-node-template/pull/81) back-port wasn't needed as it wasn't introducing a breaking or security issue. We should have waited until it arrived in an official release.

Now, talking about the **nighty** issue, I agree that your temporary fix is a solution. I prefer to push the `v2.0.0` tag as it isn't changing the code but just adds a compilation fix, pushing the tag to get this commit won't brake any existing `v2.0.0` tutorial or docs.

danforbes:
@riusricardo I still don't understand what the "huge regression" will be. We are only maintaining one version now, so I don't understand why using a `master` branch is bad. We can always change this if we switch to versioning our docs. I absolutely do not want to move a tag and would much prefer to create a new tag and update our docs to use that tag, but either way I feel like this is just additional overhead that could be reduced by using a `master` branch.

riusricardo:
Closing this as it got stalled and couldn't agree on it.

PR #732, opened 2020-10-19
remove sub0 registration annoucment

NukeManDan:
It is (unfortunately) over now 😭 

Possibly replace linking to the replays instead (VERY good btw 🥳 )?

### DIFF
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -32,9 +32,6 @@ class Index extends React.Component {
 
 		return (
 			<section>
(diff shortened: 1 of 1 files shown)


wheresaddie:
@NukeManDan feel free to update with archives url and push a PR. thanks for the quick update!

imadarai (approved):
We should leave the CSS and JS UTMs code for future announcement.  I'm ok with just removing the HTML for now.

PR #740, opened 2020-10-23
Minor structural fix to chain spec article

JoshOrndorff:
This PR change the `Extension` heading in the chain spec KB article from H2 to H3. This makes sense because the extension is part of the client spec. This also sidesteps the question that I had when reading the article

> If there are _two_ parts of the spec, like the text says, then why are there three subheadings.


riusricardo (approved):
:+1:

wheresaddie:
lgtm, thanks @JoshOrndorff

riusricardo:
I hate when this CI issue happens, it is stuck. Can you push another commit @JoshOrndorff or do you remember how we fixed it last time? :sweat_smile:

PR #747, opened 2020-10-25
add gossip

NukeManDan:
assuming gossipsub is used over other pubsub methods (not sure)

- [ ] Is the writing:
  - [ ] Clear: No jargon.

### DIFF
--- a/docs/knowledgebase/getting-started/glossary.md
+++ b/docs/knowledgebase/getting-started/glossary.md
@@ -359,6 +359,10 @@ Genesis configuration of Substrate-based chains is accomplished by way of a
 [chain specification](../../knowledgebase/integrate/chain-spec) file, which makes it easy to use a
 single Substrate codebase to underpin multiple independently configured chains.
 
+## Gossip
+
+The specific protocol ([gossipsub]https://github.com/libp2p/specs/tree/master/pubsub/gossipsub) of messaging between nodes with [libp2p](#libp2p) that substrate uses. More details on other `pubsub` messaging variations and methods are [here](https://docs.libp2p.io/introduction/what-is-libp2p/#messaging-pubsub).
+
 ## GRANDPA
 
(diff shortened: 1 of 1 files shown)


NukeManDan:
If you agree / this is correct wording, I would add links to this in the tutorials where gossip is referenced 😁

riusricardo (commented):
I think this no longer stands. We use a protocol that has some similarities to GossipSub but it doesn't has the same spec. https://github.com/paritytech/substrate/tree/master/client/network-gossip
This issue explains a little bit more about the actual approach. https://github.com/paritytech/substrate/issues/6335

wheresaddie:
thanks @riusricardo that was my guess also but I wasn't 100%, thanks for confirming

wheresaddie:
going to close this now @NukeManDan but feel free to reopen if you have further discoveries or confirmations. Thanks for all your PRs!

NukeManDan:
@wheresaddie I think we should address this before closing, and/or make this PR iterative until we define and properly inform people on the docs - @riusricardo what language would you suggest for use here, as "gossip" as a term appears a lot in the docs without a definition or reference to your links.

riusricardo:
@NukeManDan I'm not saying that **gossip** is incorrect. My point is that it isn't **gossipsub**.

As @wheresaddie said, please feel free to open it again. Sorry if you felt it as a harsh response, that wasn't the intention.
I agree that we should iterate the content, unfortunately we don't have any docs that specify the protocol since it isn't following any standard and changes depending on Polkadot's development needs.

I would suggest to do a generic description of the gossip protocol without referencing gossipsub. The part of libp2p is correct.

danforbes:
I actually asked about this term in particular when I was reworking the glossary. I believe that @joepetrowski mentioned that we tend to use it in a "general peer-to-peer networking" sense. We could potentially link out to the Wikipedia article, for instance https://en.wikipedia.org/wiki/Gossip_protocol

PR #769, opened 2020-11-05
on-chain randomness: fix link to randomness recipe

meck93:
- [x] Is the writing:
  - [x] Clear: No jargon.
- [x] If this is a new page, does the PR include the appropriate infrastructure, e.g. adding the page to a sidebar?
- [x] Do links go to rustdocs or devhub articles rather than code?
- [x] If this PR addresses an issue in the queue, have you referenced it in the description?

### DIFF
--- a/docs/knowledgebase/runtime/randomness.md
+++ b/docs/knowledgebase/runtime/randomness.md
@@ -76,7 +76,7 @@ _all_ pallet's that consume its randomness.
 
 ### Examples
 
-- Explore the [Recipe about Randomness](https://substrate.dev/recipes/3-entrees/randomness.html).
+- Explore the [Recipe about Randomness](https://substrate.dev/recipes/randomness.html).
 
 ### References
 


wheresaddie (approved):
thanks for keeping this current

burdges:
I've no idea where the text at  https://substrate.dev/recipes/randomness.html  lives but it should really be fixed to include the caveots from https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/blob/source/docs/knowledgebase/runtime/randomness.md#generating-randomness to be more in line with https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/blob/source/docs/knowledgebase/runtime/randomness.md#generating-randomness maybe:

    A naive randomness source might hash of the previous block and use it as a random seed. Such a technique has the significant disadvantage that the block author can preview the random seed, and try alternative blocks until they find one that yields a desirable outcome.  In concrete terms, they can exert n bits of influence upon this randomness by doing 2^n hashes.

    Substrate's [Randomness Collective Flip pallet](https://substrate.dev/rustdocs/v2.0.0/pallet_randomness_collective_flip/index.html) provides a convenient randomness source for testing.  It employs a low-influence function of the previous 81 block hashes that reduces each individual block hashes' influence upon the resulting randomness.  We warn that low-influence functions come with two major flaws that make it insecure when used in production.

    Randomness changes quite slowly so this only makes sense when randomness could safely be stable across a few blocks.  This is probably acceptable in tests, but almost rarely desirable in production.
    
    Any low-influence function amounts to previous block hashes "voting on randomness output bits", so although they reduce individual blocks' influence upon the randomness, but they actually increase an adversaries' influence upon the randomness.  In practice, an adversary who controls only the last few blocks could extensively influence collective flip randomness with far less hash power than required by even naive randomness.  In short, the randomness collective flip pallet is less secure for use in production than even naive randomness of using the previous block hash.

    Although it may seem harmless, you should not hash the result of the randomness provided by the collective flip pallet. Secure hash functions satisfy the [Avalance effect](https://en.wikipedia.org/wiki/Avalanche_effect) which means that each bit of input is equally likely to affect a given bit of the output. Hashing will negate the low-influence property provided by the pallet."

wheresaddie:
thanks for this comprehensive explanation @burdges .. I see @meck93 deleted the branch patch :+1:

meck93:
I restored it. If you let me know what you want to have changed additionally. I can make those changes.

PR #798, opened 2020-12-10
fix typo in tutorial

jimmychu0807:


docs/tutorials/build-permission-network/launch-network.md
```diff
@@ -129,7 +129,7 @@ Now we have 3 well known nodes.
 
 ## Add an extra connection
 
-Let's add another node, but not well known, only a "sub-node" of Charlie, which can only conneted to Charlie's well known node.
+Let's add another node, not a well-known node but a "sub-node" of Charlie, which can only connect to Charlie's well-known node.
```
riusricardo, line 132: we should stop using the term well-know here. Same as the other tutorial where we mention them.
Well-known keys are a different thing than development keys/users.

kaichaosun: Here the well-known node means that the default allowed connections in a permissioned network. 
https://github.com/paritytech/substrate/blob/master/frame/node-authorization/src/lib.rs#L108

kaichaosun (approved):
LGTM

danforbes (approved):

docs/tutorials/build-permission-network/launch-network.md
riusricardo: Yeah I agree with you @kaichaosun, but that is misleading. We shouldn't be using the same term for different definitions.
In summary it is incorrect to use well-known here.

PR #817, opened 2021-01-06
Fix #813

### LINKED ISSUE #813: Seminar Page Doesn't Render Upcoming Events
The Seminar page https://substrate.dev/seminar is not correctly rendering this recent commit with new upcoming events https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/commit/e0041327521933c416475cecd10c8ead7cf6474d

jimmychu0807:
The reason is that `<translate>` is a docusaurus specific tag. It doesn't support having nested tag inside. So 

```html
<translate>xxx<strong>xxx</strong>xxx</translate>
```

will generate an error for that snippet. 

We can do it this way:

```html
<translate>xxx</translate><strong><translate>xxx</translate></strong><translate>xxx</translate>
```

But since this part of content changes so frequently. Let's just not translate them and keep them as plain english.

### DIFF
--- a/website/pages/en/seminar.js
+++ b/website/pages/en/seminar.js
@@ -92,11 +92,8 @@ function Seminar(props) {
           <div className='row justify-content-between pt-0 faqs'>
             <div className='col-md-6 first'>
               <h4 className='mb-2'><translate>What topics are coming up?</translate></h4>
-              <p>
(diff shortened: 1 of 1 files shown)


danforbes (approved):
Thank you so much, Jimmy 🙏🏻

PR #830, opened 2021-01-19
switch to version 2.0.1 of the node template

tansaku:
v 2.0.0 did not compile for me on OSX 11.1 and rust 14.9 and this bump fixed it as described here https://stackoverflow.com/a/65756888/316729


jimmychu0807 (approved):
thanks @tansaku

PR #831, opened 2021-01-21
Fix broken links related to upgrades

hugopeixoto:
The previous link pointed to the "executor" page, which doesn't contain the "forkless runtime upgrades" and "versioning" sections anymore.
These were moved to the "runtime/upgrades" page.


jimmychu0807 (approved):
thanks @hugopeixoto

PR #832, opened 2021-01-21
Add link to contracts pallet

hugopeixoto:
I would also remove the (SEAL) parenthesis, since I don't see any references to SEAL anywhere in the documentation or the pallet page. Is this an old/internal name? Is SEAL an acronym for something?


jimmychu0807 (approved):
Thanks @hugopeixoto

jimmychu0807:
@hugopeixoto very true. Probably legacy issue.

PR #833, opened 2021-01-21
[WIP] update subkey instructions for node key

### LINKED ISSUE #791: Update private network tutorial to use key subcommand
Polkadot has a subcommand to insert keys into the keystore without having to start the node first. This eliminates the nastiness around restarting nodes after inserting grandpa keys https://github.com/paritytech/substrate/issues/7268#issuecomment-737491349

The private network tutorial should be updated accordingly.

NukeManDan:
Add `key` subcomand as primary tool for all key managment on devhub. Link to subkey as stand-alone alternative, update the KB to reflect this in [it's page](https://substrate.dev/docs/en/knowledgebase/integrate/subkey). 

- [ ] Add the option to generate a new node-id key instead of using a well known one. 
- [ ] Update the last steps to instruct on how to use a self generated and maintained key for any of the nodes
- [ ] Double check subkey for breaking updates to v2.0.1+ [install info](https://substrate.dev/docs/en/knowledgebase/integrate/subkey) still has 2.0.0 listed. 

Will close #791 (use this issue as a reference, [here is what to use with `key`](https://github.com/paritytech/substrate/issues/7268#issuecomment-737491349)

### DIFF
--- a/docs/tutorials/build-permission-network/add-node-authorization-pallet.md
+++ b/docs/tutorials/build-permission-network/add-node-authorization-pallet.md
@@ -52,9 +52,9 @@ git checkout perm-network
 You should be able to `check` the project (or `build`) without any error: 
 
 ```shell
-cd substrate-node-template/
-make init
-make check
+cd substrate-node-template #if not already there
+make init # only need to run once
+make check # run every time you want to check
 ```
 
 > If you do run into issues building, checkout
@@ -76,7 +76,7 @@ pallet-node-authorization = { default-features = false, version = '2.0.0' }
 default = ['std']
 std = [
     #--snip--
-    pallet-node-authorization/std,
+    'pallet-node-authorization/std',
     #--snip--
 ]
 ```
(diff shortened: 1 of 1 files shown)


NukeManDan:
ah it looks like it might be a bad idea to suggest doing you own key... at least in this step, as I assume the node key and account signing key need to be paired in some way? If so 
```
(
OpaquePeerId(bs58::decode("12D3KooWBmAwcd4PJNJvfV89HwE48nwkRmAgo8Vy3uQEyNNHBox2").into_vec().unwrap()),
endowed_accounts[0].clone()	
)
```

breaks. 

Are these interdependent? If so, why do we store them separately?  @riusricardo Maybe you can help clarify? 🙏🏼

NukeManDan:
Also go this error on the polkadot.js apps
```
Uncaught (in promise) Error: FATAL: Unable to initialize the API: createType(PeerId):: Cannot construct unknown type PeerId
```
- [ ] update instructions to add custom struct in the polkadot.js apps developer page

NukeManDan:
@riusricardo this is an issue: I have got both Allice, Bob and Charlie up, but no blocks are being produced. Thus I cannot get a sudo call to add Charlie to the well know nodes and anything else for that matter. 

I have the branch with my code here: https://github.com/NukeManDan/substrate-node-template/tree/permissoned-demo
image

NukeManDan:
Ah I missed this:

```
{
  "PeerId": "(Vec<u8>)"
}
```

I don't see any errors or complaints from the polkadot.js btw... So I suspect this will only break on submission of an extrinsic that relies on this? Strange! will report back

PR #834, opened 2021-01-21
use /tmp folder, not ~/tmp

NukeManDan:
So this gets cleaned up automatically on restart, instead of cluttering user's home folder 🍑


kaichaosun (approved):
👍

PR #843, opened 2021-01-28
make it clear which button to use when setting names

tansaku:
I was confused at this point in the instructions, and tried the "unsigned" and "sudo" buttons and getting various errors before working out I needed to press "signed"


jimmychu0807 (approved):
thanks @tansaku

PR #848, opened 2021-02-02
clarify this line is an update rather than an addition

tansaku:
not a big deal, but took me a few minutes to work this out from the error message


riusricardo (approved):
Thanks for making it clear @tansaku!

PR #868, opened 2021-02-24
add guidance on which file to edit

tansaku:
Assuming I'm correct that traits should be added to lib.rs, we might indicate that to the reader?


jimmychu0807 (approved):
thanks @tansaku

PR #869, opened 2021-02-24
Updating Twitter Handle

imadarai:
changing twitter handle to substrate instead of parity.io in footer (per marcomms)


riusricardo (approved):
Nice!

PR #875, opened 2021-03-03
Fall back to v2.0.1 - as contracts pallet is not in v3.0.0 yet

NukeManDan:
from the [v3 release notes](https://github.com/paritytech/substrate/releases/tag/v3.0.0)
> Contracts: Please note that the contracts pallet is not part of this release. The pallet is not yet ready and will be released separately in the coming weeks. The currently released contracts pallet is not compatible with the new FRAME, thus if you need the contracts pallet, we recommend you wait with the upgrade until it has been released, too.

So we should be using 2.0.1/0.8.1 for this until v3 (or latter release) includes it.

Related to https://github.com/paritytech/subport/issues/116

### DIFF
--- a/docs/tutorials/add-contracts-pallet.md
+++ b/docs/tutorials/add-contracts-pallet.md
@@ -8,14 +8,14 @@ blockchain. However, in the attempts to remain minimal, it does not include most
 [FRAME](../../knowledgebase/runtime/frame).
 
 This guide will show you how you can add the
-[Contracts pallet](https://substrate.dev/rustdocs/v3.0.0/pallet_contracts/) to your runtime in order to
+[Contracts pallet](https://substrate.dev/rustdocs/v2.0.1/pallet_contracts/) to your runtime in order to
 allow your blockchain to support Wasm smart contracts. You can follow similar patterns to add
 additional FRAME pallets to your runtime, however you should note that each pallet is a little
 different in terms of the specific configuration settings needed to use it correctly.
 
 ## Install the Node Template
 
-You should already have version `v3.0.0` of the
+You should already have version `v2.0.1` of the
 [Substrate Node Template](https://github.com/substrate-developer-hub/substrate-node-template)
 compiled on your computer from when you completed the
 [Create Your First Substrate Chain Tutorial](../../tutorials/create-your-first-substrate-chain/).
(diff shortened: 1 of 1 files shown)


NukeManDan:
@gnunicorn I see that there is progress in getting v3 of contracts in place, where can I get notice on this being complete? (so we can formally move the docs here to v3)

gnunicorn:
@NukeManDan it has already been released last week - https://github.com/paritytech/substrate/pull/8175

jimmychu0807:
@gnunicorn @NukeManDan but `pallet-contracts v3.0.0` is not accompanied with [Substrate v3.0.0](https://github.com/paritytech/substrate/tree/v3.0.0). In our doc, we are referring developers to this Substrate version, so we probably want to keep it consistent and use `pallet-contract v2.0.1` (which Substrate v3.0.0 is using). My two cents.

shawntabrizi:
@gnunicorn maybe you can release a minor patch which includes the new contracts pallet version?

gnunicorn:
> maybe you can release a minor patch which includes the new contracts pallet version?

Have what _include_  what?

Maybe  I should clarify: a release is a push to crates.io. The github tags are merely for reference and allowing people to look up the source code that was (almost - we've removed the dev-deps) at that point. Because crates are independent, and no crate we released depends on pallets-contracts. Crates are not always released in bundles (see 2.0.1 before), which is we are not (really) using the github releases feature: we just have too many. However, you can find the tags for these crates on the releases page:
image

What exactly do you want me to release?

NukeManDan:
Ah so as an example https://crates.io/crates/pallet-contracts-primitives is now 3.0.0 - so we can safely close this PR and instead update everything to v3/0.9 . 

> a release is a push to crates.io.

These additional crates got pushed _after_ the 3.0.0 release, right? As the release had them as "coming soon" [here](https://github.com/paritytech/subport/issues/116). In similar situations, we will have releases with pending additions, how do we go about dealing this? My mental model seems to settle on new non-breaking features being a patch or minor rev (3.0.1) 

> Since 2.0.0 we should really just point to crates.io .

IIUC - you suggest that we start using the `^major.minor.patch` dependency specification throughout for users to get all patch semver updates?

---

I know there was a session that was recorded on how semver is now used in substrate @gnunicorn - but I can't find it again, do you know of this and can share so I can review it? 🙏🏼

jimmychu0807:
@NukeManDan 
1. One that Ben talked about sem ver is on [this Substrate Seminar](https://www.crowdcast.io/e/substrate-seminar/16).

2. I think this PR can be adjusted so they all consistently points to v3.0.0. I was doing string search previously on `v2.0.x` and I missed those that do not start with a `v`. I saw some of your changes on these.

NukeManDan:
Superseded by #899 
Thanks @riusricardo 🙇🏼

PR #878, opened 2021-03-06
Upgrade the template code snippets to v3.0.0 [Tutorial: Prepare to build a dApp]

hamidra:
The tutorial in "Prepare to build a dApp" section mentions "You should already have version v3.0.0 of the Substrate Developer Hub Node Template", but it seems the code snippets are still from v2.0.0 which are not compatible with v3.0.0 and makes the pallet template fail to compile for node template v3.0.0. Apparently the frame_system::Trait module has been renamed to  frame_system::Config in v3.0.0.

- [ ] Is the writing:
  - [ ] Clear: No jargon.

### DIFF
--- a/docs/tutorials/build-a-dapp/pallet.md
+++ b/docs/tutorials/build-a-dapp/pallet.md
@@ -145,8 +145,8 @@ following code will need to be added to your `pallets/template/src/lib.rs` file:
 ```rust
 /// Configure the pallet by specifying the parameters and types on which it depends.
 pub trait Config: frame_system::Config {
-	  /// Because this pallet emits events, it depends on the runtime's definition of an event.
-	  type Event: From<Event<Self>> + Into<<Self as frame_system::Config>::Event>;
+    /// Because this pallet emits events, it depends on the runtime's definition of an event.
+    type Event: From<Event<Self>> + Into<<Self as frame_system::Config>::Event>;
 }
 ```
 
--- a/docs/tutorials/build-a-dapp/pallet.md
+++ b/docs/tutorials/build-a-dapp/pallet.md
@@ -69,7 +69,7 @@ At a high level, a FRAME pallet can be broken down into six sections:
 
 ```rust
 // 1. Imports
-use frame_support::{decl_module, decl_storage, decl_event, decl_error, dispatch, traits::Get};
+use frame_support::{decl_module, decl_storage, decl_event, decl_error, dispatch};
 use frame_system::ensure_signed;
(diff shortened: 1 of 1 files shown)


docs/tutorials/build-a-dapp/pallet.md
```diff
@@ -135,18 +135,18 @@ std = [
...
 covered in greater depth in the next tutorial, the [Add a Pallet](../add-a-pallet) tutorial.  The 
 following code will need to be added to your `pallets/template/src/lib.rs` file:
 
 ```rust
 /// Configure the pallet by specifying the parameters and types on which it depends.
-pub trait Trait: frame_system::Trait {
-    /// Because this pallet emits events, it depends on the runtime's definition of an event.
-    type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
+pub trait Config: frame_system::Config {
+	  /// Because this pallet emits events, it depends on the runtime's definition of an event.
```
shawntabrizi, line 148: be careful what happens to spacing. I think we follow that everything uses tabs

docs/tutorials/build-a-dapp/pallet.md
```diff
@@ -69,11 +69,11 @@ At a high level, a FRAME pallet can be broken down into six sections:
 
 ```rust
 // 1. Imports
-use frame_support::{decl_module, decl_storage, decl_event, decl_error, dispatch};
+use frame_support::{decl_module, decl_storage, decl_event, decl_error, dispatch, traits::Get};
```
shawntabrizi, line 72:
```suggestion
use frame_support::{decl_module, decl_storage, decl_event, decl_error, dispatch};
```

Where does the `Get` requirement come from? I think this is not right

shawntabrizi:
As a note, with Substrate v3.0 we also introduced a new Macro Syntax: https://github.com/paritytech/substrate/discussions/7788

Documentation does not yet cover this, as it might require a bunch of work to fix up, but eventually we will use this for everything. You may notice some other PRs which are updating our existing pallets to use this new macro system. It is worth your time to learn this too, in addition to the old macro.

docs/tutorials/build-a-dapp/pallet.md
hamidra: Good to know!

shawntabrizi (approved):

jimmychu0807 (approved):
@hamidra thanks for updating the tutorial to v3 compatible

PR #879, opened 2021-03-06
Knowledge base on runtime benchmarking

jimmychu0807:
There will be a companion PR on writing a tutorial of runtime benchmarking, but I will separate them in another PR. 

This is the first draft on the knowledge base. Please correct for any technical accuracy as well as any english typo / grammar mistake.

If you prefer to read in a webpage format, it is also [deployed in staging environment](https://devhub-maindocs.herokuapp.com/docs/en/knowledgebase/runtime/benchmarking). Username and password is provided in devhub Element channel.

Thanks @shawntabrizi @danforbes @NukeManDan


yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
shawntabrizi, line 1: this file probably didnt mean to exist?

docs/tutorials/runtime-benchmarking/index.md
```diff
@@ -0,0 +1,5 @@
+---
+title: Introduction
+---
+
+# Benchmark `pallet-template`
```
shawntabrizi, line 5: ??

jimmychu0807: Keeping this file as placeholder for now so no dead link and pass the CI test.

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+---
+
+## What Is Runtime Benchmarking?
+
+In Substrate, each block must be produced within a certain time limit to keep the blockchain healthy
+and functional. This is the target blocktime. So in each block production phase, Substrate can only
+take so many extrinsic calls and compute for their results. We use weight to represent how much
+computations are needed for a runtime extrinsic.
+
+Some extrinsics has more computation and has a larger weight. If this extrinsic is included, then
```
kianenigma, line 12:
```suggestion
Some extrinsics have more computation and therefore have a larger weight. If this extrinsic is included, then
```

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+there is less capacity left to include other extrinsics in the block. This is further explained in
+the [Transaction Weight chapter](../learn-substrate/weight).
+
+In Substrate, **10^12 Weight = 1 Second**, and i.e 1,000 weight = 1 nanosecond, measured on a
+referenced hardware.
```
kianenigma, line 17: need reference to what is the _reference hardware_.

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+extrinsic turns out to run with less computation, some of the estimated weight can be returned, and
+eventually causing the transaction fee returned back to user. This is further explained in the
+[Transaction Fees chapter](./fees).
+
+Now the question is how do we determine the computation time and weight of our extrinsics? There
+might also be a relationship between the time and the input or the existing state (the current
+database storage) when running the extrinsic. As any computer scientist can tell, we are getting
+into a time complexity analysis of the function.
+
+This is where Substrate Runtime Benchmarking comes in. It has a set of toolchain to help determine
```
kianenigma, line 30:
```suggestion
This is where Substrate Runtime Benchmarking comes in. It has a set of toolchains to help determine
```

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+Now with Substrate benchmarking toolchain, runtime developers can estimate the weight of their
+runtime extrinsics which translate to the transaction fee charged to the runtime users. So it is
+critical to benchmark our runtime extrinsics to measure how the extrinsic computation changes with
+repect to its input and set the proper weight functions for these extrinsics. Setting a proper
+weight function that reflects accurately on the underlying computation and storage is an important
+security safeguard in Substrate.
+
+## Features of Substrate Benchamrking Toolchain
+
+- You can configure multiple input variables to be used when benchmarking an extrinsic. For each
```
kianenigma, line 62: would be good to link to the `benchmarks!` macro do somewhere here, it as very good documentation.

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+        /* verifying final state */
+      }
+    }
+    ```
+
+- At the end, it deduces a multi-variable linear equation for the weight function on how the
+  computation time changes with respect to the change of your specified variables.
+
+- It generates a `WeightInfo` structure in Rust that can either be further tweaked, or easily
+  integrated and called in your runtime.
```
kianenigma, line 89:
```suggestion
  integrated into your runtime.
```

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+      use super::*;
+      use frame_benchmarking::{benchmarks, account, impl_benchmark_test_suite};
+      use frame_system::RawOrigin;
+
+      benchmarks!{
+        // Individual benchmark cases here
+      }
+
+      impl_benchmark_test_suite!(Pallet, crate::tests::new_test_ext(),
+        crate::tests::Test);
```
kianenigma, line 131:
```suggestion
      impl_benchmark_test_suite!(
        Pallet, 
        crate::tests::new_test_ext(),
        crate::tests::Test,
      );
```

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+
+    An example in
+    [`example/src/lib.rs`](https://github.com/paritytech/substrate/blob/master/frame/example/src/lib.rs):
+
+    ```rust
+    // This will measure the execution time of `set_dummy` for b in [1..1000] range.
+    set_dummy_benchmark {
+      let b in 1 .. 1000;
+    }: {
+      set_dummy (RawOrigin::Root, b.into());
```
kianenigma, line 159:
```suggestion
      set_dummy(RawOrigin::Root, b.into());
```

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+    The function `set_dummy` is called in the execution section, and we do not verify the result at
+    the end.
+
+    You will also see code like this:
+
+    ```rust
+    accumulate_dummy {
+      let b in 1 .. 1000;
+      let caller = account("caller", 0, 0);
+    }: _ (RawOrigin::Signed(caller), b.into())
```
kianenigma, line 177:
```suggestion
    }: _(RawOrigin::Signed(caller), b.into())
```

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+    ```rust
+    benchmarks! {
+      // -- snip
+      impl_benchmark_test_suite!(Pallet, crate::tests::new_test_ext(),
+        crate::tests::Test);
+    }
+    ```
+
+    At the last line in the `benchmarks!` macro, `impl_benchmark_test_suite!` takes three input: the
+    module name, the function that generates a test genesis storage `new_text_ext()`, and a test
```
kianenigma, line 193:
```suggestion
    module name, the function that generates a test genesis storage (i.e. `new_text_ext()`), and a test
```

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+    runtime we built when writing our test cases, and expand them into testing code.
```
kianenigma, line 194:
```suggestion
    runtime, and expand them into testing code. The test runtime can often be the same one used for testing. 
```

This is better, because some benchmarks might need special mock-ing.

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+
+    Ultimately benchmarking code is expanded as testing code, so it takes the code you write in
```
kianenigma, line 196: I find this to be a bit confusing. The whole benchmarking does indeed _not_ expand to testing code, whereas the opposite, it should be run with `--release`. The testing portion is executed in a debug release.

As noted above, I think you can rework this section maybe, just to point out that: 
1. hey, you need a mock runtime for testing your benchmarks as well
2. hey, you might be able to re-use the one from your pallet testing
3. [additional links etc]

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+    ```toml
+    [dependencies]
+    # -- snip --
+    pallet-you-named = { version = ..., default-features = false, path = ... }
+
+    [features]
+    std = [
+      # -- snip --
+      "pallet-you-named/std"
+    ]
```
kianenigma, line 215: Technically unrelated, but I agree that comparing this with std makes the learning experience better.

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+      "pallet-you-named/runtime-benchmarks"
+    ]
+    ```
+
+    Then in the node runtime, as usual, have your runtime implemements the pallet configurable trait
+    and add the pallet in `construct_runtime!`. For details, refer to [Add a Pallet to Your Runtime
+    Tutorial](../../tutorials/add-a-pallet). Then our runtime also implements
+    `frame_benchmarking::Benchmark` trait under a feature flag, and implement the
+    `dispatch_benchmark` method, as [shown in the
+    file](https://github.com/paritytech/substrate/blob/master/bin/node/runtime/src/lib.rs#L1412).
```
kianenigma, line 227:
```suggestion
    file](https://github.com/paritytech/substrate/blob/master/bin/node/runtime/src/lib.rs#L1412).
```

This link will become invalid any day as it is not bound to a commit, and only line number.

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+        // -- snip --
+
+        if batches.is_empty() { return Err("Benchmark not found for this pallet.".into()) }
+        Ok(batches)
+      }
+    }
+    ```
+
+    There seems to be a lot of code, but we are mainly setting up an environment that all
+    benchmarking code will run, such as we pass in a set of storage keys that should not be counted
```
kianenigma, line 256:
```suggestion
    benchmarking code will run. For example, we pass in a set of storage keys that should not be counted
```

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+The second output is an autogenerated `WeightInfo` implementation. This file defines weight
+functions of benchmarked extrinsics with the computed coefficient above. You can directly integrated
+this file in your extrinsics weight function or further customize them if so desired. The
+auto-generated implementation is designed to make end-to-end benchmark automation easy.
+
+To use this file, you first define a `WeightInfo` trait in
+[`frame/example/src/lib.rs`](https://github.com/paritytech/substrate/blob/master/frame/example/src/lib.rs):
+
+```rust
+pub trait WeightInfo {
```
kianenigma, line 451: hmm wait -- WeightInfo is also generated in as the output.

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,547 @@
...
+In the `frame/example/src/lib.rs`, we have:
+
+```rust
+pub mod weights;  // importing the `weights.rs` here
+
+pub mod pallet {
+  // -- snip --
+  #[pallet::call]
+  impl<T: Config> Pallet<T> {
+    #[pallet::weight(<T as pallet::Config>::WeightInfo::accumulate_dummy(0))]
```
kianenigma, line 471:
```suggestion
    #[pallet::weight(T::WeightInfo::accumulate_dummy(0))]
```

kianenigma (commented):
mostly good so far. What I would like to request is one additional example where a dispatch's weight is actually correlated with one of the parameters. in the `set_dummy`, there is this random `b` and maybe people will not learn what it _should have been_.

jimmychu0807:
thank you for the detail review @kianenigma 🙏

docs/knowledgebase/runtime/benchmarking.md
jimmychu0807: I believe by default when running benchmark as is, it doesn't generate the trait, but just the implementation. It is for Substrate now because the CI/developer is using [frame-weight-template.hbs](https://github.com/paritytech/substrate/blob/v3.0.0/.maintain/frame-weight-template.hbs) as the template to generate the `weightInfo` implementation by passing additional `--template` cli argument. 

I am not covering WeightInfo template here and just introducing the default generated `WeightInfo` impl output. 

@shawntabrizi Just want to confirm if my understanding is correct.

docs/knowledgebase/runtime/benchmarking.md
jimmychu0807: added to the footnote.

docs/knowledgebase/runtime/benchmarking.md
jimmychu0807: added on L#85

docs/knowledgebase/runtime/benchmarking.md
jimmychu0807: updated.

docs/knowledgebase/runtime/benchmarking.md
jimmychu0807: I just finish reading the [benchmark! doc](https://substrate.dev/rustdocs/v3.0.0/frame_benchmarking/macro.benchmarks.html) and want to clarify what `impl_benchmark_test_suite!` does. 

- So running `substrate benchmark` actually doesn't run the code generated by `impl_benchmark_test_suite!`?
- It generates code to be run in `cargo test`, and run with the high and low values of  complexity parameters and see if it returns an `Ok(())`?

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,567 @@
...
+      // All good, no refund.
+      Ok(().into())
+    }
+  }
+}
+```
+
+For each extrinsic, we add an attribute macro `pallet::weight` and pass in the weight implementation
+function with the same name as the extrinsic and variables that affect the weight. We can pass
+extrinsic parameters as weight function inputs to estimated the final weight.
```
jimmychu0807, line 499: @kianenigma see if this addresses your last comment.

docs/knowledgebase/runtime/benchmarking.md
kianenigma: You are correct but my question is why not cover this? I personally have found it a bit of pain to manually maintain correct signatures in `Trait WeightInfo` and the fact that it can be autogenerated, IMO, is important.

docs/knowledgebase/runtime/benchmarking.md
kianenigma: >  So running substrate benchmark actually doesn't run the code generated by impl_benchmark_test_suite!?

No, that only runs the main benchmark code. The tests are run if you go into your pallet and run `cargo test --all-features`, or if you have done: 

```
#[cfg(any(test, feature = "runtime-benchmakr))]
mod benchmarking;
``` 

Then any `cargo test` will run the benchmark code as well. 

Some pallets do this, some don't afaik. I personally like this. The only downside is that if a benchmark test takes a lot of time, it will pollute the main test suite. But this is a rare case.

kianenigma: > It generates code to be run in cargo test, and run with the high and low values of complexity parameters and see if it returns an Ok(())?

Correct. I think it generates just 3 steps per benchmark, with repeat=1, where the 3 steps are [low, mid, high] of the range.

shawntabrizi (approved):

shawntabrizi:
@jimmychu0807 i went through and made changes where i thought it made sense.

Please do one more check and then feel free to merge.

docs/knowledgebase/runtime/benchmarking.md
```diff
@@ -0,0 +1,561 @@
...
+
+However, the exact command you run will depend on your needs.
+
+> **Notes**
+>
+> Another argument introduced above is `--template`. We allow you to specify your own weight
+> template file and the benchmarking toolchain will fill in the exact numbers from the measured
+> result. This enables automating weight generation in your desired code format and integrates this
+> in your CI process.
+
```
jimmychu0807, line 342: @kianenigma added a paragraph here telling readers they can use `--template` to customize the weightinfo implementation.

jimmychu0807:
I am merging this PR.

@shawntabrizi thanks for the update.
@kianenigma I made some updates based on your feedback. See if we altogether have addressed your concerns. If not, we can address it in another PR.

PR #911, opened 2021-03-16
Create your first chain v3

NukeManDan:
Closes #881 

companion that should come first (not required) #910

PR #912, opened 2021-03-16
update links, bump version numbers to v3 in tuto cards

NukeManDan:
- fix #911 CI links 
- Bump v3 for all cards 
- card images and sorting updated

### DIFF
--- a/docs/tutorials/create-your-first-substrate-chain/background.md
+++ b/docs/tutorials/create-your-first-substrate-chain/background.md
@@ -17,9 +17,9 @@ To function, a blockchain node needs:
 - [Storage](../../knowledgebase/advanced/storage)
 - [Peer-to-peer networking](https://libp2p.io)
 - [Consensus capabilities](../../knowledgebase/advanced/consensus)
-- [Extrinsic handling](../../learn-substrate/extrinsics) (more than just transactions)
-- A [Runtime](../../knowledgebase/runtime)
-- and [more](../../index)...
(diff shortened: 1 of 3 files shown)


NukeManDan:
CI like this one - @riusricardo this should be ready to merge and it will deploy the updates 🤞🏼

riusricardo (approved):
👍

riusricardo:
We should not bump all versions of tutorials in one PR. It should be done on the same PR that updates the tutorial.
Right now it says that the permission network is on v3.0 and this is not the case. Same for the private network tutorial that still has an open PR to update it.
cc: @sacha-l (just mentioning you here to keep you on the loop)

NukeManDan:
My mistake, I should have mentioned this in the PR 🤦🏼 
Happy to have a new PR to get the right version numbers, I figured the cards would be ok to jump the gun on as we are moving fast to get them there 🚀 

LMK, as all v3 tagged issues could have the requirement instead to make sure you bump the versions correctly.
This could be in the README as a note and/or scripts helpers to remind or automatically get this correct?

sacha-l:
> We should not bump all versions of tutorials in one PR. It should be done on the same PR that updates the tutorial.
> Right now it says that the permission network is on v3.0 and this is not the case. Same for the private network tutorial that still has an open PR to update it.
> cc: @sacha-l (just mentioning you here to keep you on the loop)

@jimmychu0807 - also keeping you in the loop 🔍

PR #921, opened 2021-03-20
Shows contracts pallet page on sidebar

jimmychu0807:
@riusricardo @NukeManDan 

This page seems relevant. Do you have the context why this page is removed from the sidebar?


riusricardo (approved):
After the second migration back to the DevHub it never made it to the sidebar.
It is a good idea to have it there :+1:

PR #922, opened 2021-03-20
Remove TODO in Executor

### LINKED ISSUE #903: Remove TODO in Executor
This page is pretty old I think, and there's a leftover TODO in it. 

https://substrate.dev/docs/en/knowledgebase/advanced/executor

jimmychu0807:
Kian, I have removed the `TODO`, and removing it from showing on side navbar.

Also in my vague idea, I think the current strategy has changed a lot. Should we remove the whole content of that page?

Close #903


jimmychu0807:
@kianenigma the Runtime Executor page is updated to the best of my knowledge and your given references from https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/903#issuecomment-803260948 and https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/903#issuecomment-803260988.

Let me know if the content is technically accurate. Thanks 🙏

kianenigma:
The strategy is fine, as long as it is the same as the default of substrate. You can always override this with `--execution` as well anyhow.

docs/knowledgebase/advanced/executor.md
```diff
@@ -20,53 +20,51 @@ Before runtime execution begins, the Substrate client proposes which runtime exe
...
+- `NativeElseWasm`: Execute with the native build if possible; if it fails, then execute with WebAssembly.
 
 The default execution strategies for the different parts of the blockchain execution process are:
 
 - Syncing: `NativeElseWasm`
-- Block Import: `NativeElseWasm`
-- Block Construction: `Wasm`
+- Block Import (for non-validator): `NativeElseWasm`
+- Block Import (for validator): `AlwaysWasm`
+- Block Construction: `AlwaysWasm`
```
kianenigma, line 33: would be good to actually link to the code here?

docs/knowledgebase/advanced/executor.md
```diff
@@ -20,53 +20,51 @@ Before runtime execution begins, the Substrate client proposes which runtime exe
...
 - Other: `NativeWhenPossible`
 
+Source: [[1]](#footnote-execution-strategies-src01), [[2]](#footnote-execution-strategies-src02)
+
 ### Wasm Execution
 
 The Wasm representation of the Substrate runtime is considered the canonical runtime. Because this
 Wasm runtime is placed in the blockchain storage, the network must come to consensus about this
-binary. Thus it can be verified to be consistent across all syncing nodes.
+binary. Thus it can be verified to be consistent across all sync'ing nodes.
```
kianenigma, line 43: I think syncing is pretty common and used.

kianenigma (approved):

docs/knowledgebase/advanced/executor.md
```diff
@@ -20,53 +20,51 @@ Before runtime execution begins, the Substrate client proposes which runtime exe
 should be used. This is controlled by the execution strategy, which can be configured for the
 different parts of the blockchain execution process. The strategies are:
 
-- `NativeWhenPossible`: Execute with the native equivalent if it is compatible with the given
-wasm module; otherwise fall back to the wasm.
-- `Wasm`: Use the given wasm module..
-- `Both`: Run with both the wasm and the native variant (if compatible).
-Report any discrepancy as an error.
-- `NativeElseWasm`: First native, then if that fails or is not possible, wasm.
+- `NativeWhenPossible`: Execute with native build (if available, WebAssembly otherwise).
```
NukeManDan, line 23: I do not see these options on the node template:

```bash
# node template
./target/release/node-template --dev --tmp --execution=asd
error: 'asd' isn't a valid value for '--execution <STRATEGY>'
	[possible values: Both, Native, NativeElseWasm, Wasm]
	
	#substrate full bin build
	./target/release/substrate --dev --tmp --execution=asd
error: 'asd' isn't a valid value for '--execution <STRATEGY>'
	[possible values: Both, Native, NativeElseWasm, Wasm]
	```
	
Are these options incorrect in the docs or does the node CLI need an update (or both)?

NukeManDan (changes requested):

docs/knowledgebase/advanced/executor.md
kianenigma: I think this is the equivalent of just `Native` from the CLI.

NukeManDan: Right. I think the docs have a deprecated set of options... and what is the difference between `Native` and `NativeIfPossible` ?

NukeManDan: I assume that "when possible" (was) trying to get at the fact that WASM will be used if the runtime is upgraded? Is that still the case, or does `Native` force use of native, and ignores wasm?

jimmychu0807: The confusion arises because the CLI argument is called one thing and in `sc_client_api` it is called something similar (but not necessarily the same), and there is an underlying one-to-one mapping, [shown here](https://github.com/paritytech/substrate/blob/9b08105b8c/client/cli/src/arg_enums.rs#L120-L129)

- cli arg `Native` maps to `sc_client_api` `NativeWhenPossible`
- cli arg `Wasm` maps to `sc_client_api` `AlwaysWasm`
- cli arg `Both` maps to `sc_client_api` `Both`
- cli arg `NativeElseWasm ` maps to `sc_client_api` `NativeElseWasm`

I am using the `sc_client_api` naming in the doc.

jimmychu0807: @kianenigma There are some questions which is along the line of Dan's one. I understand executor will check the runtime version of the native and wasm runtime. If the native runtime version is the same as wasm's one, then `NativeWhenPossible` and `NativeElseWasm` strategy will choose native runtime.

But if not, and the wasm runtime version is more updated, is `NativeWhenPossible` choosing the wasm runtime? In this case, which runtime does `NativeElseWasm` strategy pick?

In what scenario do we want to use the `Both` strategy (running with both runtime)? And I assume this only application if both runtime version is the same. If native & wasm runtime version is different, there is no point in running `Both` as it will always have difference and cause error.

jimmychu0807:
@NukeManDan I have updated with Basti comment that runtime version is always respected. See if there is any more specific change you want to made. If none, will be great if you could approve the PR. Thanks 😄

NukeManDan (approved):
I still think we aught to include clatrifications WRT to the differences in the CLI and the internal API in a future update. But LGTM for now - sorry for blocking!

PR #924, opened 2021-03-20
Update event and layout on Seminar page

sacha-l:
Seminar event page update. A couple things:

- <translate> tags would break when running locally. The change that resulted in this was: "...we meet every _other_ Tuesday at 14:00UTC. "
- the Element link is mysteriously broken - not sure who to tag here to help fix this
- I switched the layout such that items under "FAQ" are a list of FAQs and are separate from the list of future events (which I moved to the right column) 

Previous v3.0 updates got mixed up here. Please ignore. Apologies for the confusion.


riusricardo (approved):

jimmychu0807:
@sacha-l The change you proposed look relevant. Why is it closed?

PR #938, opened 2021-03-25
Security file update

imadarai:
Activating Polkadot Decoded ad 
+ Adding security files as directed by Security Team.

### DIFF
--- a/.well-known/security.txt
+++ b/.well-known/security.txt
@@ -1,0 +1,5 @@
+Contact: bugbounty@parity.io
+Preferred-Languages: en
+Canonical: https://www.parity.io/.well-known/security.txt
+Policy: https://www.parity.io/bug-bounty/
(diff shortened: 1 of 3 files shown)


NukeManDan (approved):
LGTM - I am curious: what is the plan to remove this once the event is over? Just set a reminder as a new PR without the ad? (leaving the security note)

imadarai:
@NukeManDan There is a reminder set on Web Board for  April 4th and we'll switch it to registration link, instead of vote.  Happy to set a PR as well, if thats necessary.

NukeManDan:
got #940 set for this, the event is on may 20-21 yes?

imadarai:
@NukeManDan Yes.  But the voting ends on 4th and then I'm told that we'll be changing the message again.

Issue #934, opened 2021-03-23
Improvement to "Installation" instructions


DavidSSL:
So some background. I have only started with Rust at the beginning of this month and I got interested in blockchains last Sunday. Therefore that might explain why I had difficulty following the instructions on the [installation steps](https://substrate.dev/docs/en/knowledgebase/getting-started/) page.

Now the problems I had are more related to Rustup but regardless of this fact, the documentation might need some reworking in my opinion.

There is a section related to installing Rustup where we are told to run:

```console
source ~/.cargo/env
```
I didn't install Rustup via the shell script but instead used the Arch community package and as result, I didn't have the `env` file. 

However, it didn't seem to matter really as I was eventually able to go through the [creation of Substrate chain tutorial](https://substrate.dev/docs/en/tutorials/create-your-first-substrate-chain/) without any issues. I agree that yes it's my fault for not following the instructions to the letter but maybe this issue can be highlighted for people who installed Rustup the way I did. 

While the previous issue was my fault, the section about the [toolchain](https://substrate.dev/docs/en/knowledgebase/getting-started/#rust-nightly-toolchain) really slowed me down as I wasn't sure what to do because, Polkadot was mentioned while I was interested in Substrate. I simply ploughed ahead and installed the latest nightly toolchain and I had no problem going through the above referenced tutorial. This might well be due to my lack of understanding about the whole Rust ecosystem but I recall reading that knowledge of Rust would not be required for that tutorial.

Now, I would be happy to help rewrite that part of the documentation as long as it is deemed to be necessary/desirable and also if I can get some degree of support either from here or in Matrix. 

Just my 2 cents as they say...

NukeManDan:
Thanks for the feedback @DavidSSL - I would be happy to review a PR on this to improve the docs. 
- IIUC the env file is just to get the $PATH variable updated:
```shell
#!/bin/sh
# rustup shell setup
# affix colons on either side of $PATH to simplify matching
case ":${PATH}:" in
    *:"$HOME/.cargo/bin":*)
        ;;
    *)
        # Prepending path in case a system-installed rustc needs to be overridden
        export PATH="$HOME/.cargo/bin:$PATH"
        ;;
esac
```
So including a note about that for other distros could be good.

2. RE: Polkadot mentioned, it is built with substrate, and as a best practice for people actually using the template for a production product like it, it is very important to set a known good toolchain. What would you suggest (in your PR) to make that more clear?

[...]

PR #944, opened 2021-03-28
Try to improve installation instruction.

DavidSSL:
[Related to #934](https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/934)

@NukeManDan: As promised here is the change. I'm still not 100% happy with it for a variety of reasons but hopefully this helps. I've checked on WSL2 that we only need 

```console
rustup update
rustup update nightly
rustup target add wasm32-unknown-unknown --toolchain nightly
```
after which we can build "substrate-node-template"


NukeManDan (approved):
LGTM - thanks for the support 😁

PR #951, opened 2021-04-01
toolbox tab to developer drop down

tansaku:
When I run this I don't see a toolbox tab - just a Developer dropdown to get to the RPC calls - maybe the interface has changed?

Screenshot 2021-04-01 at 12 44 22

maybe the screenshot in the tutorial needs changing?


NukeManDan (approved):
I think this was legacy, now you have the right info here. Thank you!

PR #962, opened 2021-04-07
Seminar page rendering update (html & css)

jimmychu0807:
Changes mostly on seminar page ( ref to current [Seminar page](https://substrate.dev/en/seminar)):

- The `join live call >>` click is now changed to a red button clear for ppl to click (because this is the most important call-to-action button
- Refer to the **FAQ** section, you will see the text in those section seems to be covered by another layer and the `Element` click is not clickable. This is fixed by using <section/> tag in each paragraph. The issue is that the `<ul><li></li></ul>` tag seems to have special css setting that will cover the whole `<div>` section.
- a side-effect that css indentation is updated from tab to space, which is what it supposed to be based on the `.editorconfig`.


website/pages/en/seminar.js
```diff
@@ -54,11 +54,10 @@ function Seminar(props) {
...
               </translate></p>
               <section>
-                <span className='block type--fine-print'>
-                  <a href='https://www.crowdcast.io/e/substrate-seminar'><translate>
-                    join live call »
-                  </translate></a>
-                </span>
+                <a className='btn btn-lg primary-color'
+                  href='https://www.crowdcast.io/e/substrate-seminar'>
+                  <translate>Join Live Call »</translate>
```
NukeManDan, line 59:
```suggestion
                  <translate>Signup for Upcoming and Review Past Seminars</translate>
```
To make it clear that both options are there

NukeManDan (approved):
LGTM but one minor change, if you want to make it @jimmychu0807

website/pages/en/seminar.js
jimmychu0807: Thanks. I updated the wordings to `Live Call / Past Seminars` to get a balance between conveying the meaning and keeping it short.

PR #969, opened 2021-04-10
Fix image

### LINKED ISSUE #909: Runtime picture have an error
https://substrate.dev/docs/en/knowledgebase/runtime/
image

one is elections, other is elections phragmen?

jimmychu0807:
close #909


shawntabrizi:
@jimmychu0807 you lost the transparency in these photos. Probably fine overall, but not so nice when in dark mode.

NukeManDan (changes requested):
The transparent background for sure would be best. Also, should we not duplicate this file in the assets? Just point to one instance of this from the multiple places it's used?

jimmychu0807:
@NukeManDan the image has been updated to transparent background and dedup'ed.

NukeManDan (approved):
LGTM! Thanks @jimmychu0807

PR #972, opened 2021-04-13
Add fedora and update other linux packages dependencies for substrate & the FE template

grenade:
the following command / package mappings from the apt (ubuntu/debian) instructions were used:

* `sudo apt update` -> `sudo dnf update`
* `sudo apt install` -> `sudo dnf install`
  * `cmake` -> `cmake`
  * `pkg-config` -> `pkgconf-pkg-config`
  * `libssl-dev` -> `openssl-devel`
  * `git` -> `git`
  * `build-essential` -> `make automake gcc gcc-c++ kernel-devel` (see: https://unix.stackexchange.com/a/1344/39654)
  * `clang` -> `clang`
  * `libclang-dev` -> `clang-devel`
  * `curl` -> `curl`
  * `libz-dev` -> `zlib-devel`

- [x] Is the writing:
  - [x] Clear: No jargon.
- [x] If this is a new page, does the PR include the appropriate infrastructure, e.g. adding the page to a sidebar?
- [x] Do links go to rustdocs or devhub articles rather than code?
- [x] If this PR addresses an issue in the queue, have you referenced it in the description?

### DIFF
--- a/docs/knowledgebase/getting-started/index.md
+++ b/docs/knowledgebase/getting-started/index.md
@@ -43,6 +43,15 @@ export OPENSSL_LIB_DIR="/usr/lib/openssl-1.0"
 export OPENSSL_INCLUDE_DIR="/usr/include/openssl-1.0"
 ```
 
+### Fedora
+
+Run these commands from a terminal:
+
+```bash
+sudo dnf update
+sudo dnf install automake clang clang-devel cmake curl gcc gcc-c++ git kernel-devel make openssl-devel pkgconf-pkg-config zlib-devel
+```
+
 ### macOS
 
 > NOTE: The Apple M1 ARM system on a chip is not very well supported yet by rust,
--- a/docs/knowledgebase/getting-started/index.md
+++ b/docs/knowledgebase/getting-started/index.md
@@ -30,7 +30,7 @@ Use a terminal shell to execute the following commands:
 ```bash
 sudo apt update
 # May prompt for location information
-sudo apt install -y pkg-config libssl-dev git build-essential clang libclang-dev curl libz-dev
+sudo apt install -y git clang curl libssl-dev
 ```
 
 ### Arch Linux
@@ -49,7 +49,7 @@ Run these commands from a terminal:
 
 ```bash
 sudo dnf update
-sudo dnf install clang curl git
+sudo dnf install clang curl git openssl-devel
 ```
 
 ### macOS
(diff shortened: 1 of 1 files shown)


NukeManDan (approved):
I don't have a Fedora machine to test this on, but LGTM. Although, I do not think that you must have cmake & automake & make - these were deprecated in v2.0.0 that used the `makefile` flow.

docs/knowledgebase/getting-started/index.md
```diff
@@ -43,6 +43,15 @@ export OPENSSL_LIB_DIR="/usr/lib/openssl-1.0"
 export OPENSSL_INCLUDE_DIR="/usr/include/openssl-1.0"
 ```
 
+### Fedora
+
+Run these commands from a terminal:
+
+```bash
+sudo dnf update
+sudo dnf install automake clang clang-devel cmake curl gcc gcc-c++ git kernel-devel make openssl-devel pkgconf-pkg-config zlib-devel
```
NukeManDan, line 52:
```suggestion
sudo dnf install clang clang-devel curl gcc gcc-c++ git kernel-devel openssl-devel pkgconf-pkg-config zlib-devel
```
Not sure if you really need all of this either TBH... `curl` could be optional for instance. clang and gcc/g++ are both needed? cc @jimmychu0807 who should we ask about base required deps for us beyond what rust needs?

NukeManDan (changes requested):
want to get this double checked before we commit, and I will try a Fedora VM to find minimal packages.

NukeManDan:
I have installed the node and frontend template, built them, and run them successfully with the now committed deps needed for any linux distro. 

I am not sure, but openssl may not be needed, as it is already installed in all distros I have tested, I assume this is needed for `ws` connections to work. Can I get a last check on this from a core dev that these minimal deps are ok?

NukeManDan (approved):

tomaka:
Git, OpenSSL, and curl aren't needed to compile Substrate.
`pkg-config`, however, probably is.

NukeManDan:
`git` and `curl` aren't needed to compile Substrate, but they are used in our guide to install (curl rust install script, clone the templates with git)

pkg-config seems to be installed by any ubuntu / fedora base OS, so it's redundant to include here AFAIK. 

IIUC `clang` uses LLVM in a more efficient/compact build for rust, so it's preferred?

I didn't need to add `protoc` to make this work on ubuntu or fedora. I can't find this in my `bin`s either in $PATH on ubuntu... So not needed?

NukeManDan:
So from feedback, this minimal set looks OK - I will keep a lookout for other users having issues if this is incorrect. Thanks guys!

PR #987, opened 2021-04-29
Update codec.md

yjhmelody:
Outdated or incorrect description

### DIFF
--- a/docs/knowledgebase/advanced/codec.md
+++ b/docs/knowledgebase/advanced/codec.md
@@ -86,12 +86,6 @@ One or zero values of a particular type. Encoded as:
 - `0x00` if it is `None` ("empty" or "null").
 - `0x01` followed by the encoded value if it is `Some`.
 
-As an exception, in the case that the type is a boolean, then it is always one byte:
-
-- `0x00` if it is `None` ("empty" or "null").
-- `0x01` if it is the `true` value.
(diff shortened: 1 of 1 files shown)


NukeManDan:
https://github.com/paritytech/parity-scale-codec/blob/f0341dabb01aa9ff0548558abb6dcc5c31c669a1/src/codec.rs#L502-L514

Thanks for the PR @yjhmelody 😁 - although from the source above, I still see this is valid for boolean options. Am I missing something else here? Please re-open this if so.

yjhmelody:
> https://github.com/paritytech/parity-scale-codec/blob/f0341dabb01aa9ff0548558abb6dcc5c31c669a1/src/codec.rs#L502-L514
> 
> Thanks for the PR @yjhmelody 😁 - although from the source above, I still see this is valid for boolean options. Am I missing something else here? Please re-open this if so.

This can be understood as a struct or a new type, rather than having to use Option + bool to express. For the implementation of all languages, they all define a struct called OptionBoool, so this specification seems meaningless?

NukeManDan:
If you wish to make a PR that changes this is the codec, it may be accepted. But it likely also effects upstream projects that would need a companion to make this happen. If you are wanting to do this, let me know @yjhmelody - I can get you connected with the right people to start it.

PR #991, opened 2021-04-29
Updating requirements

frgomes:
Adding instructions about environment dependencies for Debian and OpenSUSE which install ``llvm-config`` and ``clang``.


NukeManDan (changes requested):
Good spot, but this should link to the installation instructions. Not introduce new ones here.
Will fix in new PR 🙇🏼

frgomes:
@NukeManDan : I've implemented the suggested changes, I guess. Thanks

NukeManDan (approved):
Thanks! 🙇🏼

Issue #965, opened 2021-04-08
Developer Documentation Needed on Proper Usage of Reference Counters


shawntabrizi:
After a review of: https://github.com/paritytech/substrate/pull/8221

The auditing team said it would be important to document how to correctly use the different reference counters so that third party teams do not make mistakes and get accounts into locked states.

A technical writer can work with me to get a better understanding of reference counters and then work to write up documentation which covers this within the DevHub or even the inline documentation.

jimmychu0807:
let me work on this next week.

jimmychu0807:
- https://github.com/paritytech/substrate/pull/8221/
- https://github.com/paritytech/substrate/pull/8220
- https://github.com/paritytech/substrate/issues/5255

@shawntabrizi I start to work on this. I went over Substrate PR #8221, see the keys are on having `{ consumers, providers, sufficients }` in `AccountInfo`. I also see the intricate logic when incrementing and decrementing each of them in the `dec_{providers, consumers, sufficients}`, and `inc_{providers, consumers, sufficients}` methods. I check and see the test cases.

But I wonder how should this be documented? Current there is almost no use of `inc_sufficients`, a few uses of `inc_providers` in the src, and moderate uses of `inc_consumers` by frame_balances, pallet_recovery. It seems to me this is some kind of internal state tracking variables. I am thinking what is a good approach of writing the doc for these three variables so runtime developers can start using them in their work?

shawntabrizi:
<img width="1079" alt="image" src="https://user-images.githubusercontent.com/1860335/117219841-144a0780-add4-11eb-91f2-71dab88b9214.png">
<img width="1079" alt="image" src="https://user-images.githubusercontent.com/1860335/117219861-1ca24280-add4-11eb-89d1-79f225719995.png">
<img width="1079" alt="image" src="https://user-images.githubusercontent.com/1860335/117219878-25931400-add4-11eb-91e6-8c5c247a2ba4.png">

PR #1006, opened 2021-05-14
Article explaining `AccountInfo` structure (including three reference counters)

jimmychu0807:
Added an article explaining Substrate account data structure, from the storage map mapping account ID to `AccountInfo`, explaining the `nonce`, three reference counters, and `AccountData` stored inside.

close: #965

@shawntabrizi Please review and see if anything need to be updated
@jakehemmerle I see your discussion with Shawn and so looping you in 😄 Please add other insights you have on this topic.


docs/knowledgebase/advanced/account-info.md
```diff
@@ -0,0 +1,149 @@
+---
+title: Account Information
+---
+
```
shawntabrizi, line 4: Would be nice to have a small preface about what the user will find on the page

docs/knowledgebase/advanced/account-info.md
```diff
@@ -0,0 +1,149 @@
...
+## `Account` StorageMap
+
+In Substrate, account information is stored in a
+[**Storage Map**](https://substrate.dev/rustdocs/v3.0.0/frame_support/storage/types/struct.StorageMap.html#impl),
+extracted and shown below:
+
+[`frame/system/src/lib.rs`](https://github.com/paritytech/substrate/blob/bcd649ffca9efc93f8b4ac1506ec8117b71e1aac/frame/system/src/lib.rs#L530-L538):
```
shawntabrizi, line 11: linking to a specific commit and lines of code probably not the best. I think just point to the file from Master

docs/knowledgebase/advanced/account-info.md
```diff
@@ -0,0 +1,149 @@
...
+pub type Account<T: Config> = StorageMap<
+  _,
+  Blake2_128Concat,
+  T::AccountId,
+  AccountInfo<T::Index, T::AccountData>,
+  ValueQuery,
+>;
+```
+
+`Account` storage map takes in five type parameters, so it has no prefix, use `Blake2_128Concat` as
```
shawntabrizi, line 26: wtf does this mean?

docs/knowledgebase/advanced/account-info.md
```diff
@@ -0,0 +1,149 @@
...
+the hash algorithm, use `T::AccountId`, the account public key as the map key, with
+`AccountInfo<T::Index, T::AccountData>` as the map value.
```
shawntabrizi, line 28: its really strange for you to say all this again when posting the struct above, idk, maybe just me.

If you want to explain the Account storage item, a simple explaination like:

"The account storage map simply maps each account id on the blockchain to an `AccountInfo` struct."

docs/knowledgebase/advanced/account-info.md
```diff
@@ -0,0 +1,149 @@
...
+
+## `AccountInfo` Structure
+
+Looking further down in the source code, we will see what `AccountInfo` is actually consisted of.
```
shawntabrizi, line 32:
```suggestion
Looking further down in the source code, we will see what `AccountInfo` actually consists of.
```

docs/knowledgebase/advanced/account-info.md
```diff
@@ -0,0 +1,149 @@
...
+  /// The number of modules that allow this account to exist for their own purposes only. The
+  /// account may not be reaped until this and `providers` are both zero.
+  pub sufficients: RefCount,
+  /// The additional data that belongs to this account. Used to store the balance(s) in a lot of
+  /// chains.
+  pub data: AccountData,
+}
+```
+
+Every account is an `AccountInfo` consisting of an `nonce` indicating number of transactions the
```
shawntabrizi, line 56:
```suggestion
Every account has an `AccountInfo` consisting of an `nonce` indicating number of transactions the
```

docs/knowledgebase/advanced/account-info.md
```diff
@@ -0,0 +1,149 @@
...
+account has sent, three reference counters, and an `AccountData` structure which contains some
+balances ([further explained below](#accountdata-structure)).
```
shawntabrizi, line 58:
```suggestion
account has sent, three reference counters, and an `AccountData` structure which can be configured
to hold different kinds of data ([further explained below](#accountdata-structure)).
```

docs/knowledgebase/advanced/account-info.md
```diff
@@ -0,0 +1,149 @@
...
+
+Let's look deeper into the three reference counters, `consumers`, `providers`, and `sufficients`.
+The reasons these counters exist because Substrate has a mechanism of slashing accounts (e.g. one
```
shawntabrizi, line 61: This is a perfect start to a new header section

shawntabrizi: > exist because Substrate has a mechanism of slashing accounts

No??

Reference counters exist to keep track of how the runtime depends on an account and vice versa... it literally tracks references of the account within the system (for example when we store data under a map controlled by an account)

docs/knowledgebase/advanced/account-info.md
```diff
@@ -0,0 +1,149 @@
...
+reason could be an account balance is below the existential deposit). If any of these reference
+counters are greater than zero, it is telling Substrate to not slash the account. For:
```
shawntabrizi, line 63: > it is telling Substrate to not slash the account.

it is telling Substrate that we should not allow the account to be destroyed, which usually means the user trying to transfer their existential deposit away

docs/knowledgebase/advanced/account-info.md
KiChjang: I think this is a misuse of "so" as a connective; here I think it means "of which" instead of "as a result".

IIRC this is also factually incorrect, having `_` as its prefix doesn't mean that it has no prefix. [During macro parsing](https://github.com/paritytech/substrate/blob/90cfb952f2e11bc6d327faf04b30dc5f4cf89df8/frame/support/procedural/src/pallet/parse/storage.rs#L383-L390), we actually expect this to be `_` at all times, which then [gets replaced by the `Ident` of the storage as the storage prefix](https://github.com/paritytech/substrate/blob/90cfb952f2e11bc6d327faf04b30dc5f4cf89df8/frame/support/procedural/src/pallet/parse/storage.rs#L82).

shawntabrizi: Yes, this is correct, there is ALWAYS a prefix

docs/knowledgebase/advanced/account-info.md
```diff
@@ -0,0 +1,149 @@
...
+
+- **`consumers` reference counter** indicates if there are any other modules depend on this account.
+An example of using this counter is in `Session` pallet when an account setting its session key(s)
+prior of becoming a validator [[1]](#ref-session-set-keys). It is also related to the `provider`
+counter, see below.
+
+- **`providers` reference counter** indicates if there are any other modules allowing this account
+to exist. Currently this counter  is incremented when a new account is created with more than the
+existential deposit to indicate this account could be used to receive consumer reference counters
+[[2]](#ref-system-created). `providers` counter is always greater than or equal to `consumers`
```
shawntabrizi, line 73: > always greater than or equal

Where are you getting this from?

docs/knowledgebase/advanced/account-info.md
```diff
@@ -0,0 +1,149 @@
...
+  on-chain for those pallets, which decrement `consumers` counter.
+
+- **`sufficients` reference counter** indicates if there are any reasons this account is
+self-sufficient to exist by itself. An example of using this counter is in `Assets` pallet when an
+account has sufficient amount of certain assets but not owning any native account balance
+[[3]](#ref-assets-new-account).
+
+Runtime developers can update these counters via `inc_consumers()`, `dec_consumers()`,
+`inc_providers()`, `dec_providers()`, `inc_sufficients()`, `dec_sufficients()` exposed by
+`frame-system`. Each increment call of a certain counter should accompanied by a corresponding
```
shawntabrizi, line 90: There are also function like `can_inc_xx` i think

docs/knowledgebase/advanced/account-info.md
```diff
@@ -0,0 +1,149 @@
...
+decrement call of the counter in an account life cycle, else it is a design bug.
+
+## `AccountData` Structure
+
+The last piece of data in `AccountInfo` is `AccountData`. It is defined in `frame-system` to be
+bound by `Member + FullCodec + Clone + Default` traits. Its actual implementation is defined in
```
shawntabrizi, line 96: > bound by `Member + FullCodec + Clone + Default`

You are just stating facts, but this is not at all helpful to the documentation.

docs/knowledgebase/advanced/account-info.md
```diff
@@ -0,0 +1,149 @@
...
+`pallet-balances`, as followed.
+
+[`frame/balances/src/lib.rs`](https://github.com/paritytech/substrate/blob/8d02bb0bfc6136f6a3c805db19f51e43090a7cd4/frame/balances/src/lib.rs#L564-L584)
```
shawntabrizi, line 99: This is not true. The configuration is in the main runtime configuration.

It just happens that most substrate chains configure it to use balances pallet, but the idea is that you can create pretty much any abstraction and stick it in there.

docs/knowledgebase/advanced/account-info.md
jimmychu0807: I am now referring to the API doc page for src code line number, and they are less likely to be outdated.

docs/knowledgebase/advanced/account-info.md
jimmychu0807: Thanks for clarification @KiChjang. Fixed.

docs/knowledgebase/advanced/account-info.md
jimmychu0807: Updated to have query functions.

jimmychu0807:
@shawntabrizi content updated based on your feedback. See if they are good now.

shawntabrizi (approved):

docs/knowledgebase/advanced/account-info.md
```diff
@@ -0,0 +1,170 @@
...
+It contains:
+
+- **Free balance** `free`. The portion of a balance that is not reserved. The free balance is
+usually what matters for most operations.
+
+- **Reserved balance** `reserved`. Reserved balance still belongs to the account holder, but is
+suspended. Reserved balance can be slashed, but only after all the free balance has been slashed.
+The total balance of an account is the sum of its free balance and reserved balance.
+
+- **Frozen balance**, splitted into `misc_frozen` and `fee_frozen`, represents balance that free
```
emostov, line 155:
```suggestion
- **Frozen balance**, split into `misc_frozen` and `fee_frozen`, represents balance that free
```

Issue #1007, opened 2021-05-14
CI not triggering updated hosted site


NukeManDan:
#804 is merged, but the page is not rendering on the site: https://substrate.dev/docs/en/knowledgebase/advanced/ss58-address-format

I _think_ this is because the link check failed. Pages should be rendered from source regardless if the CI fails this check, or IMHO any (so long as reviewer test locally that changes are as expected!)

PR #1010, opened 2021-05-17
Fixing missing links

jimmychu0807:
@NukeManDan 

Close: #1007

Clear a warning that `charset` is not proper key


NukeManDan (approved):
Thanks Jimmy!

PR #1011, opened 2021-05-18
updated weightInfo type assignment

narulasrinivas:
Update code for WeightInfo assignment as per node template v3.0.0

### DIFF
--- a/docs/tutorials/add-a-pallet/configure-a-pallet.md
+++ b/docs/tutorials/add-a-pallet/configure-a-pallet.md
@@ -103,8 +103,8 @@ impl pallet_balances::Config for Runtime {
     // The FRAME runtime system is used to track the accounts that hold balances.
     type AccountStore = System;
 
-    // No weight information is supplied to the Balances pallet by the Node Template's runtime.
-    type WeightInfo = ();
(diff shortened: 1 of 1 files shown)


docs/tutorials/add-a-pallet/configure-a-pallet.md
```diff
@@ -103,8 +103,8 @@ impl pallet_balances::Config for Runtime {
     // The FRAME runtime system is used to track the accounts that hold balances.
     type AccountStore = System;
 
-    // No weight information is supplied to the Balances pallet by the Node Template's runtime.
-    type WeightInfo = ();
+    // Weight information is supplied to the Balances pallet by the Node Template's runtime.
+    type WeightInfo = pallet_balances::weights::SubstrateWeight<Runtime>;
```
NukeManDan, line 107: I _think_ we actually intend to leave this weightless, as it's intended for testing only, although I would be for moving to using weights... @jimmychu0807 what do you think? 

@narulasrinivas - have you completed thew entire tutorial and can confirm this works as expected?

NukeManDan:
Ah I see this is a duplicate of #997 - that has more context as well. Going to close this one in favor of that one for that extra info.
Thanks for the PR @narulasrinivas 😁

PR #1015, opened 2021-05-25
Updated add-a-pallet to `v3.0.0+monthly-2021-05`

jimmychu0807:
This is the corresponding [Substrate node template solution](https://github.com/substrate-developer-hub/substrate-node-template/tree/tutorials/solutions/add-a-pallet-v3.0.0%2Bmonthly-2021-05).

PR #1018, opened 2021-05-31
Fix a typo in add-a-pallet tutorial

lmatz:
The [solution](https://github.com/substrate-developer-hub/substrate-node-template/blob/tutorials/solutions/add-a-pallet-v3/runtime/src/lib.rs#L284) is `Module` while the doc says `Pallet`.


NukeManDan:
This is about to be deprecated, `Module` is now `Pallet` everywhere moving forward. The solution will be updated in this: #1015

PR #1037, opened 2021-06-15
[setup] Updating git branch

arturgontijo:
Cloning the proper `v3.0.0+monthly-2021-05` branch as we do in the next tutorial [here](https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/blob/source/docs/tutorials/add-a-pallet/index.md).

BTW, why don't we use `master` here?

- [ ] Is the writing:
  - [ ] Clear: No jargon.

### DIFF
--- a/docs/tutorials/create-your-first-substrate-chain/setup.md
+++ b/docs/tutorials/create-your-first-substrate-chain/setup.md
@@ -23,10 +23,10 @@ You will probably need to do some set-up to prepare your computer for Substrate 
 Once the prerequisites are installed, you can use Git to clone the Substrate Developer Hub Node
 Template, which serves as a good starting point for building on Substrate.
 
-1. Clone the Node Template (version `v3.0.0`).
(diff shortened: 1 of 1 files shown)


NukeManDan (approved):
LGTM! thanks for the PR to update 🙏🏼

NukeManDan:
We don't use `master` as this may include breaking changes from time to time. Only thing worse than an outdated tutorial is a broken one 😉

arturgontijo:
Thanks @NukeManDan .

> We don't use `master` as this may include breaking changes from time to time. Only thing worse than an outdated tutorial is a broken one

What if we script all the Tutorials' steps into a single/multiple CI steps? If that makes sense I can start a draft.

arturgontijo:
Thinking a bit more here, maybe this approach would be not that simple to maintain =/
We'd need to trigger a github action in the other repos that would notify this one about inconsistencies.
Will elaborate more on this though.

PR #1041, opened 2021-06-18
Added "Past Seminars" page

### LINKED ISSUE #788: Have an index page showing all previous Substrate Seminars
- with tagging & filtering

jimmychu0807:
closes #788 

- The page looks like this now:

  **v1**: 

  ss
  
  **now (grouped by year)**: 

  Screenshot 2021-06-22 at 17 28 19

- I separated out the data in `website/data/past-seminars.json`. So @sacha-l in future you can just add a json entry there.
- If we want to make it fancier, we can add table pagination and tag filtering.


sacha-l:
This looks great @jimmychu0807, great initiative. Only thing I'm wondering is whether there's a reason you haven't included all the seminars from last year too?

al3mart:
Cool idea!

jimmychu0807:
@sacha-l 
- You are right, I want to get an opinion from the team what you guys think about adding this page
- If adding 2020 events, the list becomes too long that I need to work on pagination (which I was thinking of doing it in another PR). But let me work on the pagination part this week.

jimmychu0807:
@sacha-l @NukeManDan 
Now the past seminars are grouped by year (refer to the original PR messages). I haven't finished all the data entry of past seminars info, but let's do that together when we have time. 

I would like to have the current development merge and deploy to prod first.

website/data/past-seminars.json
```diff
@@ -0,0 +1,154 @@
...
+    "tags": ["bridge"]
+  }, {
+    "date": "2021 Mar 23",
+    "speakers": ["Raul Romanutti, Parity", "Dan Forbes, Community"],
+    "description": "The Art of Chaos",
+    "link": "https://www.crowdcast.io/e/substrate-seminar/37",
+    "tags": ["kusama"]
+  }, {
+    "date": "2021 Mar 09",
+    "speakers": ["Sasha Lansky, Parity"],
```
sacha-l, line 34:
```suggestion
    "speakers": ["Sacha Lansky, Parity"],
```

sacha-l (approved):
LGTM - caught a minor spelling error. Going forward we should keep track of the various tags. This will be useful for search etc. Thanks for this.

website/data/past-seminars.json
jimmychu0807: 😄

PR #1053, opened 2021-07-02
Adding back cumulus workshop in Tutorials Page

jimmychu0807:
@sacha-l @NukeManDan 

updated 2021-07-05:

- pending on https://github.com/substrate-developer-hub/cumulus-workshop/issues/80

I think it would be good now to add back cumulus workshop in tutorial page.

### DIFF
--- a/website/pages/en/tutorials.js
+++ b/website/pages/en/tutorials.js
@@ -199,6 +199,17 @@ class Tutorials extends React.Component {
         href: "tutorials/ink-smart-contracts-tutorial/",
       },
       {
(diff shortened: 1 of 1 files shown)


sacha-l (approved):

NukeManDan:
I think we already spoke about the desire to get a more consistent versioning  scheme for cumulus based projects so we don't have two ones to maintain. I tend to think that is best still. This github org features the cumulus workshop in the home page, that might be enough for now... I too want more parachain docs on devhub, but maybe its still immature to do so 😢

PR #1055, opened 2021-07-05
Example uses const on a non-const fn

gilescope:
(Tip off from a support request.)

https://substrate.dev/docs/en/tutorials/add-contracts-pallet/#install-the-node-template

contract_info_size isn't a const fn any more since https://github.com/paritytech/substrate/commit/30ad9184c488c8e57507c1a660bb1fd16646a0f3 (to be const it would require a nightly flag and we're trying to minimise these).


thiolliere:
it is quite hard to know which version is suppose to use the tutorial. the first line say to use the version 3.0.0 of the node template which uses substrate v3.0.0 on crates.io

And just below they say to use pallet-contract with a version which is probably not compatible with substrate 3.0.0 :
```
pallet-balances = { default-features = false, version = '3.0.0', git = 'https://github.com/paritytech/substrate.git', tag = 'monthly-2021-05'}
```

in substrate v3.0.0 there is no function named contract_info_size.
on substrate monthly-2021-05 there is a function named contract_info_size and it is not a const function.

.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
KiChjang, line 1: wat? this was never ignored before?

KiChjang (approved):

docs/en/tutorials/add-contracts-pallet.html
```diff
@@ -186,11 +186,11 @@ <h3><a class="anchor" aria-hidden="true" id="implementing-the-contract-trait"></
 
 <span class="token comment">/*** Add This Block ***/</span>
 <span class="token macro property">parameter_types!</span> <span class="token punctuation">{</span>
-    <span class="token keyword">pub</span> <span class="token keyword">const</span> <span class="token class-name">TombstoneDeposit</span><span class="token punctuation">:</span> <span class="token class-name">Balance</span> <span class="token operator">=</span> <span class="token function">deposit</span><span class="token punctuation">(</span>
+    <span class="token keyword">pub</span> <span class="token class-name">TombstoneDeposit</span><span class="token punctuation">:</span> <span class="token class-name">Balance</span> <span class="token operator">=</span> <span class="token function">deposit</span><span class="token punctuation">(</span>
```
NukeManDan, line 189: Am I missing something? This document is rendered from the MD here right?
https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/blob/source/docs/tutorials/add-contracts-pallet.md

If so, please change this file, and check it renders locally :pray:

NukeManDan (changes requested):

.gitignore
shawntabrizi: because this is the wrong branch

shawntabrizi (changes requested):
@gilescope you need to work off the `source` branch. master branch is where the final generated files are placed.

docs/en/tutorials/add-contracts-pallet.html
gilescope: Ok source branch - working with md seems a lot nicer.

gilescope:
Will reraise

PR #1060, opened 2021-07-07
Update all rustdocs links using `latest` instead of a specific version

jimmychu0807:
@sacha-l @NukeManDan 

close: #1054

### DIFF
--- a/docs/knowledgebase/advanced/account-info.md
+++ b/docs/knowledgebase/advanced/account-info.md
@@ -12,10 +12,10 @@ on-chain, it is important to understand how to use these reference counters in y
 ## `Account` StorageMap
 
 In Substrate, the entry point of account information is stored in
-[`frame-system` pallet](https://substrate.dev/rustdocs/v3.0.0-monthly-2021-05/frame_system),
+[`frame-system` pallet](https://substrate.dev/rustdocs/latest/frame_system),
 extracted below.
 
-[`frame/system/src/lib.rs`](https://substrate.dev/rustdocs/v3.0.0-monthly-2021-05/src/frame_system/lib.rs.html#530):
+[`frame/system/src/lib.rs`](https://substrate.dev/rustdocs/latest/src/frame_system/lib.rs.html#530):
 
 ```rust
 /// The full account information for a particular account ID.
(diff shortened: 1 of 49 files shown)


NukeManDan:
I disagree that we should point to a floating reference in docs. This leads to more headaches when (frequently) breaking changes occur. Rather I think being explicitly stating versions is best to avoid this. 

The rework of the devhub allows for versioning of docs themselves too, making this more apparent

sacha-l:
I think this could work but how would it be future proof for when we have versioning (re @NukeManDan's point)? It's my opinion that having the URL slug with specific versions will make it more manageable and user friendly. That being said, we could keep the scheme you propose here so long as we maintain previous numbered versions correspond to those same versions in our docs.

jimmychu0807:
@NukeManDan @sacha-l 

Currently in the [rustdocs repo](https://github.com/substrate-developer-hub/rustdocs), there are `v3.0.0-monthly-2021-07` dir and `latest`, which is a linux symbolic link pointing to `v3.0.0-monthly-2021-07` dir. With the given folder structure and symlink, you can access:

- https://substrate.dev/rustdoc/v3.0.0 (the oldest version)
- https://substrate.dev/rustdoc/v3.0.0-monthly-2021-05 (2nd oldest version)
- https://substrate.dev/rustdoc/v3.0.0-monthly-2021-07 (latest version)
- https://substrate.dev/rustdoc/latest (pointing to the same docs as `v3.0.0-monthly-2021-07`, because it is symlink'ed in the underlying file system).

What I'm suggesting is we still have the exact monthly version, but we also have the `latest` symlink pointing to the latest monthly release. In future when we update the monthly release, we can just move the `latest` symlink in the rustdocs repo, and all our rustdocs links in various docs will automatically point to the latest links without needing to manually update them.

If there are breaking changes in the underlying code and rustdoc link (change to a totally new address), we are going to update the doc content anyway and at there we will also update the rustdoc link.

By adopting this change/system, it will make our future maintenance on rustdoc links in monthly release a lot easier.

jimmychu0807:
Also, if we have a proper link checker CI tool setup, it will notify us if any of the rustdoc links are no longer working. 

Our current link checker in this repo is not working properly. And hopefully the one in devhub-v2, one imad working on, will have a properly working link checker.

PR #1061, opened 2021-07-08
Update `add-a-pallet` and merge `create-a-pallet` tutorial publishing…

### LINKED ISSUE #947: Update Create a Pallet Tutorial FRAME v2
https://substrate.dev/docs/en/tutorials/create-a-pallet/

Based on https://github.com/substrate-developer-hub/substrate-node-template/pull/160 landing from @jimmychu0807 - there will be a new tag to use for this tutorial (if not all tutorials?) for the node template to build from.

jimmychu0807:
Merge **[Write a Pallet in its Own Crate](https://substrate.dev/docs/en/tutorials/create-a-pallet/)** tutorial into **[Add a Pallet to Your Runtime](https://substrate.dev/docs/en/tutorials/add-a-pallet/)**.

close #947 

@NukeManDan @sacha-l


docs/tutorials/add-a-pallet/configure-a-pallet.md
```diff
@@ -30,42 +30,40 @@ trait for the Nicks pallet.
 
 To figure out what you need to implement for the Nicks pallet specifically, you can take a look at
 the
-[`pallet_nicks::Config` documentation](https://substrate.dev/rustdocs/v3.0.0-monthly-2021-05/pallet_nicks/trait.Config.html)
+[`pallet_nicks::Config` documentation](https://substrate.dev/rustdocs/latest/pallet_nicks/trait.Config.html)
```
NukeManDan, line 33: The use of latest and master may lead to conflicts in what the docs here have and what the source does. I would want to explicitly use a version here and make that known. This is vest practice for all of devhub I think too. What do you think?

NukeManDan (changes requested):

docs/tutorials/add-a-pallet/index.md
```diff
@@ -20,16 +20,13 @@ pallet to your FRAME runtime.
...
 [ask a question on Stack Overflow](https://stackoverflow.com/questions/tagged/substrate) and use the
-`substrate` tag or contact us on
-[Element](https://matrix.to/#/#substrate-technical:matrix.org).
+`substrate` tag or contact us on [Element](https://matrix.to/#/#substrate-technical:matrix.org).
 
 ## Install the Node Template
 
-You should already have [version `v3.0.0+monthly-2021-05` of the Node Template](https://github.com/substrate-developer-hub/substrate-node-template/tree/v3.0.0+monthly-2021-05) compiled on your computer from when
-you completed the
+You should already have [latest version of the Node Template](https://github.com/substrate-developer-hub/substrate-node-template/tree/latest) compiled on your computer from when you completed the
```
NukeManDan, line 27: We should use a consistent version

website/pages/en/tutorials.js
```diff
@@ -99,8 +99,8 @@ class Tutorials extends React.Component {
       },
       {
         img: "img/tutorials/crates.png",
-        title: <translate>Add a Pallet to Your Runtime</translate>,
-        text: <translate>Add the Nicks pallet to your Substrate node template.</translate>,
+        title: <translate>Add and Publish a Pallet of Your Runtime</translate>,
```
NukeManDan, line 102:
```suggestion
        title: <translate>Add Pallets to Your Runtime and Publish Custom Pallets</translate>,
```

website/pages/en/tutorials.js
```diff
@@ -99,8 +99,8 @@ class Tutorials extends React.Component {
...
+        text: <translate>Learn to add the Nicks pallet to your runtime and publish a pallet.</translate>,
```
NukeManDan, line 103:
```suggestion
        text: <translate>Learn to add the Nicks pallet to your runtime and publish custom pallets that others can import into their runtimes</translate>,
```

NukeManDan:
As @sacha-l was mentioning in #947 , I think a HTG fits  publishing a pallet better, but this is OK for now to consolidate the tutorials

docs/tutorials/add-a-pallet/configure-a-pallet.md
jimmychu0807: The use of `latest` tag is going to be different from master branch. `master` branch is more like a development branch in our case where we have features and fixes applied but not yet released.

`latest` tag is always referring to the latest RELEASED VERSION. And in our doc, we can then refer to this latest tagged version. From then on, we don't need to update the release number in the doc every time when a monthly release is made. And when a monthly release is made, we can just move the latest tag up to the latest monthly version.

docs/tutorials/add-a-pallet/index.md
jimmychu0807: same as above

docs/tutorials/add-a-pallet/configure-a-pallet.md
NukeManDan: Ok now I understand where you are coming from, Although we need to scan for breaking changes in any case, and there may be deprecated docs that are missed just because people didn't realize moving the latest tag/branch should have docs changes. It's best practice to _never_ change a tag from what I hear, as it's assumed static. If people ask about something, we need to have them clarify _when_ they looked at the docs, not what docs they looked at (or ask for the commit)

jimmychu0807: The rule about tag to not move is generally correct. I think the exception is on the `latest` tag (or the similar setting). The name from it implies it is going to be floating and like a symlink in an OS, pointing to the exact version.

sacha-l:
[Here's](https://substrate.dev/substrate-how-to-guides/docs/basics/publish-pallet) the link to the HTG btw.

PR #1079, opened 2021-07-31
Make the tutorial work from a remote host

t1m0t:
I've added a Troubleshooting section to explain how to access the Front-End Template http://localhost:8000 without having the error `Error Connecting to Substrate { "isTrusted": true }` when the node is running on a remote host.
The issue has been raised here: https://github.com/substrate-developer-hub/substrate-front-end-template/issues/186

- [ ] Is the writing:
  - [ ] Clear: No jargon.

### DIFF
--- a/docs/tutorials/create-your-first-substrate-chain/interact.md
+++ b/docs/tutorials/create-your-first-substrate-chain/interact.md
@@ -144,6 +144,14 @@ the results of asynchronous operations. If you have already used the Front-End T
 balance transfer as described above, you should see an event for the transfer in the Event component
 next to the Pallet Interactor.
 
+## Troubleshooting
+
+In case of this error `Error Connecting to Substrate { "isTrusted": true }` and if you run the node as a remote host, you can create ssh local port forwarding to access the Front-End Template at http://localhost:8000/:
+```bash
+# Enable local port forwarding on port 9944
+ssh -L 9944:127.0.0.1:9944 <remote user>@<remote host ip> -N -f
+```
+
 ## Next Steps
 
 🎉 Congratulations!!! 🎉


docs/tutorials/create-your-first-substrate-chain/interact.md
```diff
@@ -144,6 +144,14 @@ the results of asynchronous operations. If you have already used the Front-End T
 balance transfer as described above, you should see an event for the transfer in the Event component
 next to the Pallet Interactor.
 
+## Troubleshooting
+
+In case of this error `Error Connecting to Substrate { "isTrusted": true }` and if you run the node as a remote host, you can create ssh local port forwarding to access the Front-End Template at http://localhost:8000/:
```
NukeManDan, line 149:
```suggestion
If you get `Error Connecting to Substrate { "isTrusted": true }` 
when connecting to a node on remote host, for _testing only_ you can create 
an `ssh` local port forwarding to access tyour node from a local Front-End Template 
at http://localhost:8000/:
```

NukeManDan (approved):

docs/tutorials/create-your-first-substrate-chain/interact.md
NukeManDan: this woks, but also linking here (using the right syntax) to the notes that already exist here is needed: Well, actually I would have expect from you guys and that I found myself is adding ssh local forwarding. It is actually suggested in the private network part. And the command would be as follow: ssh -L 9944:127.0.0.1:9944 <remote user>@<remote ip> -N -f
Thanks anyway :)

NukeManDan: a mention of the _right_ was with a wss cert woulds be needed (I think(

NukeManDan (changes requested):
Note: this error is thrown for _any_ issue connecting to the node. So the instructions should make that clear

docs/tutorials/create-your-first-substrate-chain/interact.md
t1m0t: @NukeManDan  Thank you for the review. This issue was solved by doing ssh local forward. If you have any other solution that works also, please do mention it.

PR #1080, opened 2021-08-01
Renaming improvements 

sacha-l:
- Move `try-runtime` article under "Integrate"
- Rename article "Why are there no Hash collections in sp_std?"  to "Hash Collections"


NukeManDan:
My first thought on "toolchains" is the rust toolchain 😅  - not sure what the overall impression you want to send with the title of the section.

sacha-l:
> My first thought on "toolchains" is the rust toolchain  - not sure what the overall impression you want to send with the title of the section. I think "Integrations" might be a _bit_ misleading, but implies the thing I _think_ it should contain: `tooling to use with substrate nodes` It's not about developing a substrate runtime/node, but _integrating_ other applications with your node. 

This change is the result of [this PR](https://github.com/paritytechmarcomms/substrate-docs/pull/11/files). I can't remember where this discussion originally occurred but I had voiced that "Integrate" is misleading and we concluded that ["Toolchains"](https://en.wikipedia.org/wiki/Toolchain#:~:text=In%20software%2C%20a%20toolchain%20is,a%20set%20of%20related%20programs.) is more appropriate (I think it was you that suggested it too!). I seem to remember my original suggestion was "Integration and Tooling".

@jimmychu0807 , would love to hear your thoughts here so we can all align. 

> Also unclear why you think hashmaps should move from "Advanced" to this section with that framing of the section's theme.

The change I made here was only a title change for the hash-collections article. The path remains:
`docs/knowledgebase/advanced/no-hash-collections.md `

NukeManDan (approved):

NukeManDan:
Got ya. Well worst case here is that we get feedback from the builders that the title is confusing. And this may break external links (to things like subkey)

Have you done a serve of our other repos and the Polkadot wiki for this? We need to before this lands IMHO.

sacha-l:
> Got ya. Well worst case here is that we get feedback from the builders that the title is confusing. And this may break external links (to things like subkey)
> 
> Have you done a serve of our other repos and the Polkadot wiki for this? We need to before this lands IMHO.

I decided to keep the "Integrate" menu name for now. It's not worth the overhaul. We'll be migrating to our new devhub soon enough.

docs/index.md
```diff
@@ -58,7 +58,7 @@ Substrate allows developers to make choices between technical freedom and ease o
 
 - Refer to the developer documentation for the [FRAME system for runtime development](knowledgebase/runtime).
 - Learn how to create rich client applications for any Substrate-based chain by using the
-  [Polkadot-JS](knowledgebase/integrate/polkadot-js) family of libraries.
+  [Polkadot-JS](knowledgebase/toolchains/polkadot-js) family of libraries.
```
NukeManDan, line 61: Should we keep the same path structure? To not break external links.

NukeManDan (changes requested):

NukeManDan (approved):

PR #1093, opened 2021-08-15
add additional behaviour at installation errors

lorenz-haenggi-jarowa:
some errors occurred and can be solved while removing --fast option (as in substrate bug 3890 mentioned)

- [x] Is the writing:
  - [x] Clear: No jargon.

### DIFF
--- a/docs/knowledgebase/getting-started/index.md
+++ b/docs/knowledgebase/getting-started/index.md
@@ -93,6 +93,12 @@ For most users you can run our script to automate the steps listed below:
 curl https://getsubstrate.io -sSf | bash -s -- --fast
 ```
 
+if you receive an error as "Skipped cargo install of 'substrate' and 'subkey'" just run the script above without --fast (solved bug see - [substrate bug 3890](https://github.com/paritytech/substrate/issues/3890))
+```bash
+curl https://getsubstrate.io -sSf | bash -s --
+```
+
+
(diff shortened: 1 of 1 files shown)


NukeManDan:
Thanks for the PR! Although this is _expected_ behavior. Almost no one wants to install the substrate node binary, and subkey we install separately: https://substrate.dev/docs/en/knowledgebase/integrate/subkey

So IMHO this isn't a helpful note. Happy to be wrong though, asking for more opinions 🙏🏼

sacha-l:
Thanks for your contribution here @lorenz-haenggi-jarowa. I agree with @NukeManDan though that the `--fast` flag is the intended behaviour here. A couple of other things I'll add in case others are curious about this topic:

- the "_Skipped cargo install 'substrate' and 'subkey'_ " is not actually an error. It's the `echo` of the [build script](https://getsubstrate.io/) under the if/else clause for specifying the `--fast flag`
- this script is also used [in the Subkey article](https://substrate.dev/docs/en/knowledgebase/integrate/subkey) with the specific command `cargo install --force subkey --git https://github.com/paritytech/substrate --version 2.0.1 --locked` to specify a version to install `subkey`. 
- the script also gives the option for developers to install the [`substrate` binary](https://github.com/paritytech/substrate/tree/master/bin/node/cli), which is most likely useful for more advanced developers who have a good reason to do so, but certainly not required for developers just getting started

Hope that makes sense!

PR #1107, opened 2021-09-04
Update `nicks-pallet` parameter types for nicks pallet ver4.0

whereistejas:
Correcting incorrect parameter type.


gilescope:
LGTM. Let's get this merged.

NukeManDan (approved):
I will test locally and revert if there is issues here, but yes this is the right way to do it from what I see :) thanks!

Issue #1122, opened 2021-09-21
Issues with Add Contracts Pallet Tutorial


shawntabrizi:
<img width="1144" alt="image" src="https://user-images.githubusercontent.com/1860335/134092504-5aa2c248-bff0-4eef-9ff4-f183e4a05112.png">

it worked with me with the feedback and the following tutorial updates can help.

* DenyAll should be imported.
* Make it more clear about the expected version. Although it is some how mentioned but not clear.
* In the other sections that include import mention all tags should match.
* DenyAll is deprecated. Update the tutorial with alternative.
* Add a section at the end with the common errors.

PR #1118, opened 2021-09-15
Fix #1117 and clarify about UI refresh.

tifrel:
Addresses #1117. Also contains clarifies that the Apps/Explorer might need a refresh to before upgrades show/are usable, as I got quite confused about seeing version 102 of the blockchain, but still getting 500 pUnits as ExistentialDeposit, when I should have gotten 100. Refreshing fixed that.

Issue #1117, opened 2021-09-13
Forkless Upgrade tutorial. Error in code snippets


claudiociardelli:
The code to be added to construct_runtime! in the "Forkless Upgrade" tutorial is incorrect. Causes an error inthe cargo build.
page https://substrate.dev/docs/en/tutorials/forkless-upgrade/sudo-upgrade

Line 
    Scheduler: pallet_scheduler::{Module, Call, Storage, Event<T>},
should be
    Scheduler: pallet_scheduler::{Call, Storage, Event<T>},

removing "Module" compiles correctly.

claudiociardelli:
In fact Line
Scheduler: pallet_scheduler::{Module, Call, Storage, Event},
should be
Scheduler: pallet_scheduler::{Pallet, Call, Storage, Event},

"Module" should be replaced with "Pallet"

NukeManDan:
This depends on your substrate version. v3.0.0 is used here, where Module is correct. 
What version are you using?

claudiociardelli:
You are right. Looking at cargo tree most of the dependencies are 4.0.0.

tifrel:
Made a PR which also clarifies that the browser UI needs (at least for me) a refresh, before I can access the Scheduler pallet extrinsics, and another one to see the updated existentialDeposit.

Issue #1094, opened 2021-08-16
Update Documentation -- pallet-nicks version number 


srinivasnarula:
Current documentation of [Nicks Pallet](https://substrate.dev/docs/en/tutorials/add-a-pallet/import-a-pallet#importing-the-nicks-pallet-crate) has version = '3.0.0' `. When I use version = `'3.0.0' ` and did the dependency check. I'm getting below errors.

`cargo.toml`
![Screenshot from 2021-08-16 21-16-07](https://user-images.githubusercontent.com/2491903/129627886-d99d4297-94b7-4506-8695-39832cf717b7.png)

`console output`
![Screenshot from 2021-08-16 21-16-14](https://user-images.githubusercontent.com/2491903/129627895-d30d338e-5dbf-4f41-ab6b-198939709fea.png)

When I update version=`4.0.0` and tag=`monthly-2021-08`, things started working.
![Screenshot from 2021-08-16 22-10-13](https://user-images.githubusercontent.com/2491903/129629889-22f0b6aa-dd7f-4a29-a8fe-6255a4d989dd.png)

I'm happy to submit PR for this change.

PR #1123, opened 2021-09-21
Update tutorial's cargo dependencies tags

pedromfedricci:
For readers that are going through the tutorials after [#232](https://github.com/substrate-developer-hub/substrate-node-template/pull/234) landed, they are likely to have issues with conflicting/old cargo dependencies such as #1122 , #1117, and #1094, if they inadvertently copy the Cargo.toml snips that were not updated.
This PR updates the tags for the tutorials, leaving only **forkless-upgrade** tutorial out as there is an open PR updating it already: #1118.
ty.

### DIFF
--- a/docs/tutorials/add-a-pallet/import-a-pallet.md
+++ b/docs/tutorials/add-a-pallet/import-a-pallet.md
@@ -42,10 +42,12 @@ your runtime has. For example, it depends on the [Balances pallet](https://subst
 
 **`runtime/Cargo.toml`**
 
-```TOML
-[dependencies]
-#--snip--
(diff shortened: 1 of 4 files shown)


NukeManDan (approved):
LGTM, although this *might* break some of the code snippets, specifically migrating from v3.0.0 -> the latest monthly.

NukeManDan:
@sacha-l I would vote to merge here, and subsequently look for breaking changes in the tutorials that this may cause. I _think_ that as @pedromfedricci points out, using anything other than the version in the `first chain` instructions is fraught with peril for using mismatching types over code snippet issues (that we can patch quickly if users run into them before us)

If you agree, let's squish it :grin:

sacha-l (approved):
LGTM - thanks for these changes @pedromfedricci 😎

PR #1128, opened 2021-09-22
Lg/contract pallet tutorial fixes

lsgunnlsgunn:
- [X] Are the audience and objective of the document clear? E.g. a document for developers that should teach them about transaction fees.
- [x] Is the writing: (it's a step in the right direction 
  - [-] Clear: No jargon.
  - [-] Precise: No ambiguous meanings.
  - [-] Concise: Free of superfluous detail.
- [-] Does it follow our style guide?
- [ N/A] If this is a new page, does the PR include the appropriate infrastructure, e.g. adding the page to a sidebar?
- [-] Build the page ($ cd website && yarn start). Does it render properly? E.g. no funny lists or formatting.
- [ ] Do links go to rustdocs or devhub articles rather than code? (I didn't test all of the links)
- [X] If this PR addresses errors reported on Element

PR #1131, opened 2021-09-23
Formatting nit picks

NukeManDan:
Patch for #1128 
(An attempt to) use hard tabs in rust code blocks
Also uses the .editorconfig specified linting, with max char (soft limit) 100 chars.

Can't seem to get Prettier to play nice with 2 space tabs in MD in the context of code blocks within a numbered list... I guess 3 is OK?

### DIFF
--- a/docs/tutorials/add-contracts-pallet.md
+++ b/docs/tutorials/add-contracts-pallet.md
@@ -1,26 +1,40 @@
 ---
-title: "Add the Contracts pallet to a runtime"
+title: "Add the Contracts Pallet to a Runtime"
 ---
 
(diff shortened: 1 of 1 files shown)


lsgunnlsgunn (approved):
Thanks.

NukeManDan:
Thank you, Lisa! Stoked to have you on the team!!! :tada:

sacha-l:
@NukeManDan - this is great but I think it would make more sense to channel these efforts to our new docs rather than applying them to our current docs which will be archived in less than a month. Perhaps also tie this in with our contribution guidelines ? As in, specifications for using Prettier and formatting guidelines for contributors. This will be an important addition ! 😃

PR #1134, opened 2021-09-26
Update the first part of the Transaction fees and weights section

lsgunnlsgunn:
- [ ] Is the writing:
  - [ ] Clear: No jargon.

### DIFF
--- a/docs/knowledgebase/runtime/fees.md
+++ b/docs/knowledgebase/runtime/fees.md
@@ -1,113 +1,138 @@
 ---
-title: Transaction Fees and Weights
+title: Transaction fees and weights
 ---
 
-When a block author constructs a block, it must limit the block's execution time. A block body
-consists of a series of [extrinsics](../learn-substrate/extrinsics). Since the resources needed to
-execute an extrinsic can vary, Substrate provides a flexible mechanism called "weights" to
-characterize the _time_ it takes to execute an extrinsic. To be economically sustainable and to
-limit spam, some transactions --- primarily those dispatched by users --- require a fee prior to
-transaction execution.
+Because the resources available to a blockchain are limited, it’s important to manage how blocks consume them. 
+The resources that need to be managed include:
 
-Although an extrinsic's weight is only one component of the fee charged to its sender, it is
-recommended to understand the [weight system](../learn-substrate/weight) before reading this
-document.
+* Memory usage
+* Storage input and output
+* Computation
+* Transaction and block size
+* State database size
 
-## Fee Calculation
+Substrate provides block authors with several ways to manage access to resources and to prevent individual components of the chain from consuming too much of any single resource. 
+Two of the most important mechanisms available to block authors are **weights** and **transaction fees**.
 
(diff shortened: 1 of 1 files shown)


docs/knowledgebase/runtime/fees.md
```diff
@@ -1,113 +1,138 @@
 ---
-title: Transaction Fees and Weights
+title: Transaction fees and weights
```
shawntabrizi, line 2: It is not proper to use title case?

https://titlecaseconverter.com/

docs/knowledgebase/runtime/fees.md
```diff
@@ -383,7 +408,7 @@ impl<T: Get<Perquintill>> Convert<Fixed128, Fixed128> for TargetedFeeAdjustment<
 }
 ```
 
-## Next Steps
+## Next ~~W~~teps
```
shawntabrizi, line 411: ^ I dont get it :)

docs/knowledgebase/runtime/fees.md
```diff
@@ -1,113 +1,138 @@
...
+To convert the weight to Currency, the runtime must define a `WeightToFee` struct that implements a conversion function, `Convert<Weight,Balance>`. 
 
-final_fee = fee * NextFeeMultiplier
-```
+Note that the extrinsic sender is charged the inclusion
+fee before the extrinsic is invoked. The fee is deducted from the sender's balance even if the transaction fails upon execution. 
+
+### Accounts with an insufficient balance
+
+If an account does not have a sufficient balance to pay the inclusion fee and remain alive—that is, enough to pay the inclusion fee and maintain the minimum **existential deposit**—then no fee is deducted and the
```
timorl, line 91: I don't think the comment about remaining alive is true – some transactions make sure an account does not disappear as the result of them, but not all do. The latter can result in an account dropping below the existential deposit, and then disappearing.

Due to payments happening before execution, it is even possible this happens and then a refund is issued that would be enough to keep that account alive, but that cannot happen so the refund is just not issued at all – strange corner case that I saw handled in the code.

timorl (commented):
Thanks a lot for this, looks accurate and readable. <3 I just found one thing that I don't think is correct.

You might also want to at least space your `-`s, or even better use `–`s for separating thoughts.

docs/knowledgebase/runtime/fees.md
timorl: The most obvious example is the difference between `transfer` (https://github.com/paritytech/substrate/blob/136120674703bf5541ee2368ebb330d803b61cce/frame/balances/src/lib.rs#L279) and `transfer_keep_alive` (https://github.com/paritytech/substrate/blob/136120674703bf5541ee2368ebb330d803b61cce/frame/balances/src/lib.rs#L390). The former might make an account go below the existence threshold, the latter ensures it doesn't. The system enforces nothing in this regard as far as I can tell.

On the other hand I'm actually not quite sure how this interacts with fees, since these are transfers and that's what ensuring continued existence means here. So, uh, what you wrote before might actually be right? <_<"

PR #1141, opened 2021-10-01
banner style updates

imadarai:
Banner style updates to make it more prominent per Marcomms.

 
Screenshot 2021-10-01 at 14 43 43


NukeManDan (approved):
(untested locally) LGTM!

